PR #85, opened 2022-04-25
Fix/cross dw and load vars

### LINKED ISSUE #47: Re-evaluate the case/when statement to define model_type
Is it common to have a reports folder?

What should our catch-all be? "Other" vs. "unidentified"?

_Originally posted by @wasilaq in https://github.com/dbt-labs/pro-serv-dag-auditing/pull/43#discussion_r825117501_

b-per:
- Load vars to models to be used with joins
- Refactor code to use those models
- Update models to use syntax that works cross DW

I have tested the code on Snowflake, BigQuery and Postgres and it worked

Closes #47


b-per:
Closing to recreate it and trigger a CI check.

PR #90, opened 2022-04-29
remove cruft

dave-connors-3:
This is a:
- [X] bug fix PR with no breaking changes
- [ ] new functionality

## Description & motivation
The logic in this model didn't make any sense! this update excludes _all resources_ from the primary package from the `stg_all_resources` model. This means that any place a user queries the DAG of their project beyond the base layer (related: could exclude earlier if we wanted), it will only represent their own project and any other imported package. 

image

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [x] Snowflake


graciegoheen (approved):
As discussed, we'll simplify the logic for now and re-visit this once we've had internal feedback.

PR #96, opened 2022-05-05
Fix/cross dw and load vars

b-per:
Replica of #85 which had some comments.
I am creating this new PR in order to trigger the CI check.


b-per:
Did not trigger Circle CI run for all commits as expected.
Closing in favor of #100

PR #100, opened 2022-05-09
Fix/cross dw , add load vars and enable circle ci

b-per:
Replica of https://github.com/dbt-labs/dbt-project-evaluator/pull/85 which had some comments.
I am creating this new PR in order to trigger the CI check.


dbt_project.yml
```diff
@@ -23,12 +23,23 @@ clean-targets:         # directories to be removed by `dbt clean`
...
+    marts:
+      dag:
+        +materialized: table
+    staging:
+      graph:
+        intermediate:
+          int_direct_relationships:
+            # required for BigQuery and Redshift for performance/memory reasons
+            +materialized: "{{ 'table' if target.type in ['bigquery', 'redshift'] else 'view' }}"
+        stg_naming_convention_folders:
```
graciegoheen, line 35: Should we move `stg_naming_convention_folders` and `stg_naming_convention_prefixes` into a separate folder since they're not related to the graph object? Maybe something like `vars`?

macros/loop_vars.sql
```diff
@@ -0,0 +1,24 @@
+{% macro loop_vars(vars) %}
```
graciegoheen, line 1: Just out of curiosity is there an advantage to using this method vs. our `select_from_values` macro?

models/marts/structure/fct_staging_directories.sql
```diff
@@ -23,8 +23,8 @@ staging_models as (
         child,
         child_file_path,
         parent_source_name,
-        left(child_file_path, len(child_file_path) - charindex('/',reverse(child_file_path)) + 1) as child_directory_path,
-        {{ dbt_utils.split_part('child_file_path', "'/'", -1) }} as child_file_name
+        replace(child_file_path, child || '.sql', '') as child_directory_path,
+        regexp_replace(child_file_path,'.*/','') as child_file_name
```
graciegoheen, line 27: genius - thanks for making this way less complicated!

models/staging/graph/stg_all_graph_resources.sql
```diff
@@ -10,61 +10,90 @@ with unioned as (
...
+
+final as (
+    select 
+        *, 
+        case 
+            when resource_type in ('test', 'source', 'metric', 'exposure', 'seed') then null
+            -- by default we will define the model type based on its prefix in the case prefix and folder types are different
+            else coalesce(model_type_prefix, model_type_folder, 'other') 
+        end as model_type,
+        row_number() over (partition by resource_id order by position_folder desc) as position_rk
```
graciegoheen, line 92: What does `rk` stand for?

models/staging/graph/stg_naming_convention_folders.sql
```diff
@@ -0,0 +1,11 @@
+{% set vars_prefix = ['staging_folder_name','intermediate_folder_name','marts_folder_name'] %}
```
graciegoheen, line 1: Do we want to add the ability to add other folders? `other_folder_name`

macros/loop_vars.sql
b-per: `select_from_values` is a macro used to efficiently load quite a lot of data into different data warehouses, because doing many `select x union all select y` behaves poorly after hundreds/thousands of statements.

on the other hand, `loop_vars` is looping through vars and just loading with the typical `select x union all select y` because performance is not an issue

So the 2 macros are just doing different things, hence why we need 2 😄

models/marts/structure/fct_staging_directories.sql
b-per: To be honest, my first goal was not too make it less complex but just to make it work across all DWs 🤣 

There is an issue currently with `dbt_utils.split_part` where having a negative index only works in Snowflake (the PR to fix it is [here](https://github.com/dbt-labs/dbt-utils/pull/559) but not merged yet)

models/staging/graph/stg_all_graph_resources.sql
b-per: `rk` is short for `rank`. I just want one row in case my joins are 1-n instead of 1-1. I am filtering it in my last query (can't use qualify in all data warehouses)


Do you want me to rename the column?

models/staging/graph/stg_naming_convention_folders.sql
b-per: My feeling would be that we start as-is, and mention in the docs what someone would have to do if she/he wanted to add more folders.

(another option would be to create a variable that lists all the different variables (incep-variable-tion) so a person would not have to update the model, they would just have to update the variable of variable, but it doesn't feel like this is a must have for the first release)

dbt_project.yml
b-per: @graciegoheen , I looked at it and I think that there is a bit more we have to do:
- create a `structure` folder under `staging` for those models (same name as the folder under `marts`)
- change `stg_all_graph_resources` to `int_all_graph_resources` and move it under `marts/.../int`
- move models from `staging/graph/intermediate` to `marts/dag/intermediate` (or similar)

I'd rather get the current PR over the line (there are quite a few changes in here) and deal with the structure changes in another issue. Good with you?

dave-connors-3: Agree with this comment, but also agree it's not blocking! I also think since these files are mostly different in terms of configs we might want to use the `{{ config() }}` in each model?

models/marts/structure/fct_staging_directories.sql
```diff
@@ -23,8 +23,8 @@ staging_models as (
         child,
         child_file_path,
         parent_source_name,
-        left(child_file_path, len(child_file_path) - charindex('/',reverse(child_file_path)) + 1) as child_directory_path,
-        {{ dbt_utils.split_part('child_file_path', "'/'", -1) }} as child_file_name
+        replace(child_file_path, child || '.sql', '') as child_directory_path,
```
dave-connors-3, line 26: small suggestion to use the [dbt_util for replace](https://github.com/dbt-labs/dbt-utils/blob/main/macros/cross_db_utils/replace.sql) -- may just help future proof this a bit?

models/staging/graph/graph.yml
```diff
@@ -10,3 +10,9 @@ macros:
       - name: values
         type: list
         description: The list of values to be inserted in the view/table. Each item of the list has the format "('value_col1', 'value_col2', 'value_col3')"
+  - name: loop_vars
```
dave-connors-3, line 13: ty for documenting these macros

models/staging/graph/stg_all_graph_resources.sql
graciegoheen: Got it! Thanks for the explanation!

dbt_project.yml
graciegoheen: Sounds good! Thanks for making a list!

graciegoheen (approved):
💯

dave-connors-3 (approved):

b-per:
Finally!

models/marts/structure/fct_staging_directories.sql
b-per: Changed 😄

PR #113, opened 2022-05-24
Add information on the variables available

b-per:
This is a:
- [x] doc PR
- [ ] bug fix PR with no breaking changes
- [ ] new functionality

## Link to Issue 

Closes #102

PR #110, opened 2022-05-23
add docs to all models that need it

dave-connors-3:
This is a:
- [x] bug fix PR with no breaking changes
- [ ] new functionality

## Description & motivation
this adds docs to all the models from our pacakge that didn't have them

I also 
- discovered that a couple staging and fct models were also missing tests! added those
- saw that i had an extra line left over from correcting an earlier merge conflict, sneaking that in there too

## Integration Test Screenshot

image

## Checklist
- [ ] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake
- [ ] I have updated the README.md (if applicable)
- [x] I have added tests & descriptions to my models (and macros if applicable)
- [ ] I have added an entry to CHANGELOG.md

PR #120, opened 2022-05-26
Feature/inception variable for layers

### LINKED ISSUE #92: How to include additional model types (base, report, etc.)
Maybe this is a future improvement, but it would be neat to dynamically get the list of variables so that users could add _additional_ model types. Thinking something like get all of the variables that end with `_folder_name` (and then `_prefixes` for `stg_naming_convention_prefixes`)... that way, someone could theoretically add variables like `report_folder_name` and `report_prefixes` and they would be appropriate picked up by these models.

_Originally posted by @graciegoheen in https://github.com/dbt-labs/dbt-project-evaluator/pull/85#discussion_r859025945_

b-per:
This is a:
- [x] refactor of existing functionality
- [ ] new functionality

## Link to Issue 

Closes #92 
We also mentioned the need for this functionality in #113 and #110

## Description & motivation

Allow users to modify the list/name of layers in their dbt project using variables instead of requiring them to update/deactivate/create models from the package to reflect their setup.

image

image

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/README.md
+++ b/README.md
@@ -614,8 +614,9 @@ vars:
 ```
 
 - naming conventions variables
-  - all the `xxx_folder_name` variables are used to parametrize the name of the folders for the `staging`, `intermediate` and `marts` layers of your DAG. Those layers are the ones we recommend in our [dbt Labs Style Guide](https://github.com/dbt-labs/corp/blob/main/dbt_style_guide.md). If you want to setup more layers or different ones, you could create new variables, override the model `stg_naming_convention_folders.sql` with the new list of variables and deactivate the model from the package in `dbt_project.yml`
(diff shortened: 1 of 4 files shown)


graciegoheen:
Does `list_layers` map directly to our idea of `model_type` in `int_all_graph_resources`? If so should we change the name of one of them so that the match? Either:
- `model_types` and `model_type`
- `list_layers` and `layer`

graciegoheen (changes requested):
Once the above has been addressed we're GOLDEN!

b-per:
Updated the branch following the comments from the team

PR #125, opened 2022-06-07
Updating get_metrics macro

callum-mcdata:
This is a:
- [X] small fix PR with no breaking changes
- [ ] new functionality

## Description & motivation

The `get_metrics` macro is super useful but there were a few missing components. Namely `dimensions` and `filters`. It also included the `model.name` aspect which doesn't exist in the metric node. It only contains the base model name (including ref) which raises a whole bunch of string parsing errors so I got rid of it. 

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [X] Snowflake

### DIFF
--- a/macros/unpack/get_metrics.sql
+++ b/macros/unpack/get_metrics.sql
@@ -12,14 +12,21 @@
             '{{ node.name }}',
             '{{ node.resource_type }}',
             '{{ node.original_file_path }}',
-            cast('{{ is_not_empty_string(node.description) | trim }}' as boolean),
+            cast('{{ dbt_project_evaluator.is_not_empty_string(node.description) | trim }}' as boolean),
             '{{ node.type }}',
-            '{{ node.model.identifier }}',
             '{{ node.label }}',
             '{{ node.sql }}',
             '{{ node.timestamp }}',
-            '{{ node.package_name }}'
-
+            '{{ node.package_name }}',
+            '{{ node.dimensions|join(', ') }}',
+            {% if node.filters|length %}
+              {% for filt in node.filters %}
+                '{{ filt.field }}'||'{{ filt.operator }}'||'''{{ filt.value }}'''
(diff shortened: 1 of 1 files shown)


macros/unpack/get_metrics.sql
```diff
@@ -36,13 +43,14 @@
               'file_path', 
               'is_described', 
               'metric_type', 
-              'model',
```
dave-connors-3, line 39: any reason to remove this field? I think this is failing in a downstream model! See the CircleCI error, i think it's referenced in `int_all_graph_resources`

callum-mcdata: AH - got it, was wondering where that failure was coming from. I guess this depends on whether we want to include blank fields for this. We're going to be adding `expression` metrics soon that will be dependent on parent metrics instead of models and so the model information in the node will be empty.

If its got relevance to downstream models used in this package, we can definitely add it back in and just populate a blank string value.

callum-mcdata: I've re-added it back in to see if it passes integration tests. The `model` attribute of the `metric` node doesn't have an identifier aspect so this will always return a blank value.

macros/unpack/get_metrics.sql
```diff
@@ -22,8 +22,16 @@
...
             '{{ node.sql }}',
             '{{ node.timestamp }}',
-            '{{ node.package_name }}'
-
+            '{{ node.package_name }}',
+            '{{ node.dimensions|join(' - ') }}',
+            {% if node.filters|length %}
+              {% for filt in node.filters %}
+                '{{ filt.field }}'||'{{ filt.operator }}'||'''{{ filt.value }}'''
+                {% if not loop.last %}|| ' - '{% else %}{% endif %}
```
b-per, line 30: Super tiny comment, but do we need `{% else %}` here?

b-per (approved):
🥖

Issue #75, opened 2022-04-19
Determine materialization strategy based on # of children / parent nodes


graciegoheen:
Is this possible?

view vs. table

graciegoheen:
Related to https://github.com/dbt-labs/dbt-project-evaluator/pull/160

PR #129, opened 2022-06-10
Feature/75 models materializations

b-per:
This is a:
- [x] new functionality

## Link to Issue 
Related to #75 but does not recommend the materialization

## Description & motivation

Those models/tests check chains of "non-physically-materialized" (e.g. views of ephemeral models, often simplified to just views in the comments below), highlighting potential cases for moving models to tables/incremental.

When long chains of views are executed it is often possible that the first physically-materialized model has poor performance or that tests on downstream models fail or take longer because of performance reason.

The output of the new fact table looks like below. Someone could sort the table by `distance` to figure out the longest chains in their project.
image

The changes cover quite a few things that I would be keen to get feedback on:
- the logic itself
  - I modified the macro `recursive_dag` to accept a parameter so I could reuse it to identify chains of views. I was thinking of leveraging directly the model`int_all_dag_relationships` and trying to remove the paths where there was a model in the middle, but it turned out to be much more complex that generating a new DAG that only looks at views/ephemeral
  - the new `int_all_dag_relationships__chained_views.sql` contains only ephemeral/views except for the most downstream model, which can be a physically materialized one (table, incremental, snapshot)
  - I have created a new folder for `materialization` checks knowing that we might have more later on
  - I have limited the `fct` to only a few columns but could add more
- testing the logic
  - I am not testing the path itself because it is not easy/possible to load arrays with `seeds`. I believe that the current test is good enough
  - I moved one of the "test project" models from ephemeral to table to be able to check that tables are excluded from the "chained view" logic. But the drawback is that the table actually needs to exist when running integration tests. I have decided to create an `on-run-start` hook in the `integration_tests` project to handle the creation of the table

Still TODO (but waiting for feedback before):
- [ ] Update README, CHANGELOG and docs
- [ ] Update the description of the macro `recursive_dag`
- [ ] Add a test on the model `int_all_dag_relationships__chained_views`

Example with BigQuery:
image

## Checklist
    - [x] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [x] Snowflake


models/marts/core/int_direct_relationships.sql
```diff
@@ -52,11 +53,21 @@ direct_relationships as (
         on all_graph_resources.resource_id = exposures.resource_id
 ),
 
+-- getting the materializationof the parent model to be able to track back chains of views
+direct_relationships_with_materialization as (
```
dave-connors-3, line 57: I am wondering if we can avoid using this CTE to map attributes to parents and children, and instead rely on the recursive logic to to this? Feels duplicative to do a self join here as well as in the next downstream model.

integration_tests/seeds/materialization/materialization_seed.yml
```diff
@@ -0,0 +1,16 @@
+version: 2
+
+seeds:
+  - name: tst_fct_chained_views
```
dave-connors-3, line 4: do you mind updating the seed name to have the `test_` prefix for consistency?

integration_tests/seeds/materialization/tst_fct_chained_views.csv
```diff
@@ -0,0 +1,14 @@
...
+model.dbt_project_evaluator_integration_tests.stg_model_1,stg_model_1,model.dbt_project_evaluator_integration_tests.int_model_4,int_model_4,1
+model.dbt_project_evaluator_integration_tests.stg_model_1,stg_model_1,model.dbt_project_evaluator_integration_tests.int_model_5,int_model_5,1
+model.dbt_project_evaluator_integration_tests.stg_model_1,stg_model_1,model.dbt_project_evaluator_integration_tests.int_model_5,int_model_5,2
+model.dbt_project_evaluator_integration_tests.stg_model_1,stg_model_1,model.dbt_project_evaluator_integration_tests.dim_model_7,dim_model_7,2
+model.dbt_project_evaluator_integration_tests.stg_model_1,stg_model_1,model.dbt_project_evaluator_integration_tests.dim_model_7,dim_model_7,3
+model.dbt_project_evaluator_integration_tests.stg_model_2,stg_model_2,model.dbt_project_evaluator_integration_tests.stg_model_4,stg_model_4,1
+model.dbt_project_evaluator_integration_tests.stg_model_2,stg_model_2,model.dbt_project_evaluator_integration_tests.dim_model_7,dim_model_7,2
+model.dbt_project_evaluator_integration_tests.stg_model_3,stg_model_3,model.dbt_project_evaluator_integration_tests.fct_model_6,fct_model_6,1
```
dave-connors-3, line 13: i think this is a view --> table relationship. Is this intended to be included?

dave-connors-3 (changes requested):
I like this a lot! I think we may be able to simplify some of the logic here if we 

1. take out the final CTE from `int_direct_relationships, and only include the parent materialization
2. update the filter to check this materialized value rather than the `direct_parent_materialized` value
3. update the recursive logic to include mapping materializations

integration_tests/seeds/materialization/tst_fct_chained_views.csv
b-per: This is intentional yes, because my original idea was to show what model is getting impacted by the chain of views. In that case, `fct_model_6` is potentially impacted by the fact that there are `x` views upstream

integration_tests/seeds/materialization/materialization_seed.yml
b-per: Done!

models/marts/core/int_direct_relationships.sql
b-per: I had a look and I think that this join will then need to happen anyway in the recursive logic. 
With the current logic, I would believe that having it as a filter at the "import CTE" level allows us to make the recursive DAG for chained_views more performant.

Wouldn't having the logic here make it easier for you for your `ephemeral/view` check for exposures? In that case you don't have to query the fill DAG but just the direct relationships.

integration_tests/seeds/materialization/tst_fct_chained_views.csv
b-per: Changing the name of the fact table to make it clearer!

models/marts/core/int_direct_relationships.sql
graciegoheen: I agree with Dave here, that it would be better to avoid duplicating this join. If we move this into the recursive logic, we can just `select * from int_all_dag_relationships where distance = 1`. That is the way that we have written the queries for many other fct_ models (_even_ in cases where it's only looking at direct relationships).

graciegoheen: I agree with Dave here, that it would be better to avoid duplicating this join. If we move this into the recursive logic, we can just `select * from int_all_dag_relationships where distance = 1`. That is the way that we have written the queries for many other fct_ models (_even_ in cases where it's only looking at direct relationships).

macros/recursive_dag.sql
```diff
@@ -1,14 +1,25 @@
...
     select  
         *
     from {{ ref('int_direct_relationships') }}
+
+    {% if not filter_views %}
+    -- when not filtering on views, we want all nodes, except tests
     where resource_type <> 'test'
+    {% else %}
+    -- when filtering on views, we only want models and consider relationships that are
+    -- either views, ephemerals or sources
```
graciegoheen, line 17: The naming here is a bit confusing.. why for `filter_views` do we include `view`s, `ephemeral`s, and `source`s?

integration_tests/seeds/materialization/test_fct_chained_views_dependencies.csv
```diff
@@ -0,0 +1,14 @@
+parent_id,parent,child_id,child,distance
+model.dbt_project_evaluator_integration_tests.fct_model_9,fct_model_9,model.dbt_project_evaluator_integration_tests.stg_model_5,stg_model_5,1
```
graciegoheen, line 2: Can we remove the `parent_id` and `child_id` and just use parent and child for readability in the seed file?

models/marts/core/int_all_dag_relationships__chained_views.sql
```diff
@@ -0,0 +1,9 @@
+-- creates a cte called all_relationships that will either use "with recursive" or loops depending on the DW
+{{ dbt_project_evaluator.recursive_dag(filter_views=true) }}
+
+select 
+    * 
+from all_relationships
+where 
+    distance <> 0
```
graciegoheen, line 8: Can we move this filter downstream, just so this model is consistent with `int_all_dag_relationships`?

graciegoheen (changes requested):
Left some comments above!

integration_tests/seeds/materialization/test_fct_chained_views_dependencies.csv
dave-connors-3: +1 !

dave-connors-3: is the path worth adding here as well? so the user would know the exact path in case there are multiple paths between the models returned?

macros/recursive_dag.sql
dave-connors-3: maybe `filter_materialized` could be closer? i guess in dbt parlance materialized would also apply to views

models/marts/core/int_direct_relationships.sql
dave-connors-3: @b-per and I chatted live about this earlier this week and I _think_ this might be the only way to achieve benoit's intended output -- might be worht having a larger convo next week!

dave-connors-3 (commented):
agree with Grace's comments here -- couple small naming questions, and I think we should maybe chat about the recursion logic one more time before merging!

integration_tests/seeds/materialization/test_fct_chained_views_dependencies.csv
graciegoheen: Maybe worth adding to the model, but not necessary to add to the seed!

PR #134, opened 2022-06-14
Add support for Databricks/Spark

b-per:
This is a:
- [ ] new functionality

## Description & motivation
This PR delivers initial support for databricks/spark.
Noteworthy:
- the package now loads the external `spark_utils` package
- a few macros were created directly in this package but could move to `spark_utils` (e.g. the array ones)
- limited support for `listagg` was added for databricks. It doesn't use all the parameters and should not move to `spark_utils` in its current form but it works for this project
- I removed the column casting for seeds which was creating problems and only kept it for selected columns where BigQuery was raising issues

Keen to get some feedback.

If it is ok to merge, we will also need to:
- update the docs and the project for BQ specific config and wording (e.g. the variable `max_depth_bigquery` which doesn't apply to only BigQuery but also Databricks)
- add to the README which adapters have been tested
- add Databricks to the CI/CD check ?

When running on databricks:
image

## Checklist
    - [x] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [x] Snowflake
    - [x] Databricks

### DIFF
--- a/integration_tests/seeds/dag/dag_seeds.yml
+++ b/integration_tests/seeds/dag/dag_seeds.yml
@@ -2,27 +2,18 @@ version: 2
 
 seeds:
   - name: test_fct_multiple_sources_joined
-    config:
-      column_types:
-        CHILD: varchar
-        SOURCE_PARENTS: varchar
     tests:
       - dbt_utils.equality:
+          name: equality_fct_multiple_sources_joined
           compare_model: ref('fct_multiple_sources_joined')
           compare_columns:
             - child
             - source_parents
 
   - name: test_fct_direct_join_to_source
-    config:
-      column_types:
-        PARENT: varchar
-        PARENT_RESOURCE_TYPE: varchar
-        CHILD: varchar
-        CHILD_RESOURCE_TYPE: varchar
-        DISTANCE: int
     tests:
       - dbt_utils.equality:
+          name: equality_fct_direct_join_to_source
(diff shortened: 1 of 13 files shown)


integration_tests/seeds/dag/dag_seeds.yml
```diff
@@ -2,27 +2,18 @@ version: 2
...
-    config:
-      column_types:
-        PARENT: varchar
-        PARENT_RESOURCE_TYPE: varchar
-        CHILD: varchar
-        CHILD_RESOURCE_TYPE: varchar
-        DISTANCE: int
     tests:
       - dbt_utils.equality:
+          name: equality_fct_direct_join_to_source
```
dave-connors-3, line 16: purely out of curiosity, what are these configs necessary?

dave-connors-3 (approved):
no real comments! thank you so much for turning this around so quickly!

I'll let @graciegoheen comment more specifically on some of that list behavior, but it seems fine to me. 

I think adding docs is the right next step! I am also pretty curious how much work it would be to add databricks to our CircleCI runs 🤔

integration_tests/seeds/dag/dag_seeds.yml
b-per: I have added those to avoid our tests being called `equality_test_fct_direct_join_to_source_col1_col2_col3_...fct_direct_join_to_source...`, making it easier to read and look for compiled SQL.

PR #144, opened 2022-06-19
Change list column_names in select_from_values to column that can be …

VasiliiSurov:
…a column_name or tuple (column_name, data_type)

This is a:
- [x] bug fix PR with no breaking changes
- [ ] new functionality

## Link to Issue 

Closes #137 

## Description & motivation

Changing column_names in `select_from_values` to columns, which can be either column_name (by default) or tuple (column_name, data type). If `values` is empty these data types are used to define correct type in empty model, by default column name will be `string`.

## Integration Test Screenshot

```
(dbt) sova@pop-os:~/IdeaProjects/dbt-project-evaluator/integration_tests$ dbt build
01:28:08  Running with dbt=1.1.1
01:28:08  Found 45 models, 48 tests, 0 snapshots, 0 analyses, 466 macros, 0 operations, 16 seed files, 5 sources, 1 exposure, 1 metric
01:28:08  
01:28:09  Concurrency: 8 threads (target='dev')
01:28:09  
01:28:09  1 of 95 START table model snadbox.stg_exposure_relationships ................... [RUN]
01:28:09  2 of 95 START table model snadbox.stg_exposures ................................ [RUN]
01:28:09  3 of 95 START table model snadbox.stg_metric_relationships ..................... [RUN]
01:28:09  4 of 95 START table model snadbox.stg_metrics .................................. [RUN]
01:28:09  5 of 95 START view model snadbox.stg_naming_convention_folders ................. [RUN]
01:28:09  6 of 95 START view model snadbox.stg_naming_convention_prefixes ................ [RUN]
01:28:09  7 of 95 START table model snadbox.stg_node_relationships ....................... [RUN]
01:28:09  8 of 95 START table model snadbox.stg_nodes .................................... [RUN]
01:28:10  5 of 95 OK created view model snadbox.stg_naming_convention_folders ............ [OK in 1.19s]
01:28:10  9 of 95 START table model snadbox.stg_sources .................................. [RUN]
01:28:10  6 of 95 OK created view model snadbox.stg_naming_convention_prefixes ........... [OK in 1.23s]
01:28:10  10 of 95 START seed file snadbox.test_fct_direct_join_to_source ................ [RUN]
01:28:12  1 of 95 OK created table model snadbox.stg_exposure_relationships .............. [CREATE TABLE (2.0 rows, 0 processed) in 2.57s]
01:28:12  11 of 95 START seed file snadbox.test_fct_documentation_coverage ............... [RUN]
[...]

PR #147, opened 2022-06-21
Do not merge - Copy of #144 for triggering CI

b-per:
Do not merge - Copy of #144 for triggering CI


dave-connors-3:
I think we can close this now that we have updated Circle CI to run on forked PRs!

PR #152, opened 2022-06-22
Run in CI without exposures or metrics

b-per:
This is a:
- [ ] new functionality
- [x] an update to the CI flow

This change should also check if the first models run fine when the dbt project of the users doesn't have any exposure or metric. Most of the errors raised so far are due to having no exposure/metric.

It is expected that this check will fail right now

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -20,3 +20,8 @@ fi
 
 dbt deps --target $1 || exit 1
 dbt build -x --target $1 --full-refresh || exit 1
+
+# test without exposures or metrics
+rm models/marts/exposures.yml
+rm models/marts/metrics.yml
+dbt run -x --target $1 --select +int_direct_relationships +int_all_graph_resources || exit 1


dave-connors-3:
This looks good to me -- my only question is how normal removing files during CI runs is 😅

dbeatty10:
### TLDR
- There's a way to use dbt projects for testing, but wrap them up and execute them with `pytest`.
- Level of difficulty: medium (not super easy, not super hard)

### Details
I'm working on a "hello world" project that introduces how to use `pytest` for CI. There's enough tricky little things that it's hard to get going without a complete working example. I'll let you know when it's ready so you can assess if it will be useful for cases like this. Regardless, we'll definitely want to switch over to `pytest` eventually for the CI part (thankfully there's a one-to-one correspondence).

b-per:
Thanks for the feedback. In the meantime, is it OK to merge this so the work we are doing on handling no exposure/metric can be tested automatically?

dave-connors-3:
@b-per yeah let's merge for the timebeing!

dave-connors-3 (approved):
🦞

b-per:
Vamos!

PR #159, opened 2022-06-27
add metrics to int_direct_relationships

### LINKED ISSUE #139: Metrics lineage is not reported
Any reason why we are joining the exposures parents but not the metrics in the code below?

It looks like something I missed (the CTE `direct_metrics_relationships` is created but not used) but I wanted to confirm that this is not something we implemented like this by design.

https://github.com/dbt-labs/dbt-project-evaluator/blob/12b1c2cc531e7ffe48f8040dbed1ad25374fa5a7/models/marts/core/int_direct_relationships.sql#L17-L52

graciegoheen:
This is a:
- [ ] new functionality

## Link to Issue 

Closes #139 

## Description & motivation

Added metrics to `int_direct_relationships`

Screen Shot 2022-06-27 at 11 18 28 AM

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake

### DIFF
--- a/models/marts/core/int_direct_relationships.sql
+++ b/models/marts/core/int_direct_relationships.sql
@@ -42,6 +42,7 @@ direct_relationships as (
         CASE 
             WHEN all_graph_resources.resource_type = 'source' THEN NULL
             WHEN all_graph_resources.resource_type = 'exposure' THEN exposures.direct_parent_id
+            WHEN all_graph_resources.resource_type = 'metric' THEN metrics.direct_parent_id
             WHEN all_graph_resources.resource_type IN ('model', 'snapshot', 'test') THEN models.direct_parent_id
             ELSE NULL
         END AS direct_parent_id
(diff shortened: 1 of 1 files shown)


dave-connors-3 (approved):
🦞

b-per (approved):
🥖 

Doesn't it feel odd though that a bug didn't raise a failing test and that we didn't create a new test that would have failed without the fix?

graciegoheen:
@b-per I suppose we're not really doing anything yet with the metrics, so they don't yet show up in any of the integration tests. I could add a new integration test to check the output of `int_direct_relationships`, `int_all_graph_resources`, and `int_all_dag_relationships` (though the final would be a bit tricky since we can't add arrays into seed files)?

b-per:
My question was kind of rhetorical and I don't know if we should solve it now...
Putting a test on all those models just for this change looks like an overkill

Maybe we will test those later with other use cases

PR #160, opened 2022-06-27
added new fct_chained_views_dependencies model - option 2

graciegoheen:
This is a:
- [x] new functionality

An alternative solution to #129 

## Description & motivation

- Created new model `fct_chained_views_dependencies` which returns the models that depend on a chain of views/ephemeral models
- Added new field `is_dependent_on_chain_of_views` to `int_all_dag_relationships`

Screen Shot 2022-06-27 at 2 57 02 PM

## Checklist
    - [x] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake

### DIFF
--- a/README.md
+++ b/README.md
@@ -724,7 +724,7 @@ You can set your own threshold for chained views by overriding the `chained_view
 
 #### Reason to Flag
 
-You may experience a long runtime for a model when it is build on top of a long chain of "non-physically-materialized" models (views and ephemerals). Most of this runtime is actually *compilation* time. In the example above, nothing is really computed until you get to `table_1`. At which point, it is going to run the query within `view_4`, which will then have to run the query within `view_3`, which will then have the run the query within `view_2`, which will then have to run the query within `view_1`. These will all be running at the same time, which creates a long runtime for `table_3` and potential storage issues if the data is sufficiently large. 
+You may experience a long runtime for a model when it is build on top of a long chain of "non-physically-materialized" models (views and ephemerals). In the example above, nothing is really computed until you get to `table_1`. At which point, it is going to run the query within `view_4`, which will then have to run the query within `view_3`, which will then have the run the query within `view_2`, which will then have to run the query within `view_1`. These will all be running at the same time, which creates a long runtime for `table_3`. 
 
 #### How to Remediate
 
--- a/README.md
+++ b/README.md
@@ -709,13 +709,13 @@ A new yml file should be created in `marts/` which contains all tests and docume
 
 ## Performance
 ### Chained View Dependencies
+#### Model
 
 `fct_chained_views_dependencies` ([source](models/marts/performance/fct_chained_views_dependencies.sql)) contains models that are dependent on chains of "non-physically-materialized" models (views and ephemerals), highlighting potential cases for improving performance by switching the materialization of model(s) within the chain to table or incremental. 
 
 This model will raise a `warn` error on a `dbt build` or `dbt test` if the `distance` between a given `parent` and `child` is greater than 5.
 You can set your own threshold by overriding the `chained_views_threshold` variable. [See overriding variables section.](#overriding-variables)
 
-#### Model
 #### Reason to Flag
 #### How to Remediate
 #### Example
@@ -779,6 +779,7 @@ vars:
 ```
 
 #### Naming Convention Variables:
+
 | variable    | description | default     |
 | ----------- | ----------- | ----------- |
 | `model_types` | a list of the different types of models that define the layers of your dbt project | staging, intermediate, marts, other |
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -744,6 +748,13 @@ A new yml file should be created in `marts/` which contains all tests and docume
         ├── staging.yml
 ```
 
+## Performance
```
dave-connors-3, line 751: assuming these are just placeholders?

dave-connors-3 (approved):
🦞 -- a few small comments, and i think we'll need some docs, but I think this is my preferred method for tracking materializations. I think relying on the recursive logic to map out the parent-child relationships feels like the right move here!

README.md
graciegoheen: Yep! Was waiting to add documentation until we confirmed which strategy we wanted to move forward with

dbt_project.yml
```diff
@@ -69,5 +72,8 @@ vars:
   marts_prefixes: ['fct_', 'dim_']
   other_prefixes: ['rpt_']
 
+  # -- Performance variables --
```
graciegoheen, line 75: @dave-connors-3 @b-per I'm curious if either of you have any thoughts as to what this threshold should be / if this is the proper way to be testing this model?

graciegoheen: This is different than the way `models_fanout_threshold` is handled - but my assumption was that in this case, someone might want to see all models dependent on chains of views when querying this model (rather than just the ones that surpass their threshold). Let me know what you think

README.md
graciegoheen: (And once we've determined the appropriate value for `chained_views_threshold` so I can give a clear example)

models/marts/performance/fct_chained_views_dependencies.sql
```diff
@@ -0,0 +1,21 @@
...
+    where distance <> 0
+),
+
+final as (
+    select
+        parent,
+        child, -- the model with potentially long run time / compilation time, improve performance by breaking the upstream chain of views
+        distance,
+        path,
+        is_dependent_on_chain_of_views
```
b-per, line 14: We could remove this column from the `fct` no? (we are already filtering on it)

dbt_project.yml
b-per: I like to show all of them and then just have the test on the `fct` failing, like you did.

README.md
```diff
@@ -786,7 +818,23 @@ vars:
...
 
+#### Performance Variables:
+
+| variable    | description | default     |
+| ----------- | ----------- | ----------- |
+| `chained_views_threshold` | threshold for unacceptable length of chain of views for `fct_chained_views_dependencies` | 4 |
+
+```yml
+# dbt_project.yml
+# set your chained views threshold to 8 instead of 5
```
b-per, line 829: We talk about the default being 4 but then say "change from 5". I guess it is just a typo.

README.md
```diff
@@ -703,6 +707,33 @@ A new yml file should be created in `marts/` which contains all tests and docume
...
+
+#### Example
+
+`table_1` depends on a chain of 4 views (`view_1`, `view_2`, `view_3`, and `view_4`).
+
+<img width="500" alt="dag of chain of 4 views, then a table" src="https://user-images.githubusercontent.com/53586774/176299679-39028eb1-f9e3-492a-bdb7-b72d9d7958b7.png">
+
+#### Reason to Flag
+
+You may experience a long runtime for a model when it is build on top of a long chain of "non-physically-materialized" models (views and ephemerals). Most of this runtime is actually *compilation* time. In the example above, nothing is really computed until you get to `table_1`. At which point, it is going to run the query within `view_4`, which will then have to run the query within `view_3`, which will then have the run the query within `view_2`, which will then have to run the query within `view_1`. These will all be running at the same time, which creates a long runtime for `table_3` and potential storage issues if the data is sufficiently large.
```
b-per, line 727: I'd be keen to work a bit on the description here. Are we sure that the time is "*compilation*" time and that there might be "storage issues".

b-per (changes requested):
Thanks Grace!

Overall, I think that this PR is the one that should be merged instead of my original one.

I just added a couple of comments. The main one is around the wording of why chained views are potentially creating performance issues.

README.md
graciegoheen: typo! I'll fix it!

models/marts/performance/fct_chained_views_dependencies.sql
graciegoheen: sure, i'll remove it

README.md
graciegoheen: There doesn't seem to be a ton of dbt documentation on this topic (was pulling from Pat's [notion writeup](https://www.notion.so/dbtlabs/Model-and-Query-Optimization-dea6ec77096349658e38a55553ebbf13)) - feel free to propose a change with a better description!

b-per: In my mind, making it a bit less specific looks better:

You may experience a long runtime for a model when it is built on top of a long chain of "non-physically-materialized" models (views and ephemerals). In the example above, nothing is really computed until you get to `table_1`. At which point, it is going to run the query within `view_4`, which will then have to run the query within `view_3`, which will then have the run the query within `view_2`, which will then have to run the query within `view_1`. These will all be running at the same time, which creates a long runtime for `table_1`.

graciegoheen: I like it! I'll make that change :)

b-per (approved):
🥖

PR #132, opened 2022-06-13
Feature/exposure materializations

dave-connors-3:
This is a:
- [ ] bug fix PR with no breaking changes
- [x] new functionality

## Link to Issue 
closes #27 

## Description & motivation
This PR adds a test to check that the models referenced directly by an exposure are properly materialized in the warehouse as either a `table` or `incremental` model. 

I am not attached to the name of the model, happy for feedback here. I also did not include DAG screenshots because I was unsure how to represent materializations in a DAG. 

There's some pretty decent overlap here with @b-per's open PR on nested views -- I elected _not_ to add the direct parent materialization column into `int_all_dag_relationships` and instead let the recursive logic take care of that join. I understand that this is required to filter out non-tables before the recursive logic, but I am wondering if it's possible for us to rely only on the recursive logic instead of doing similar operations in two places

## Integration Test Screenshot

image

## Checklist
- [ ] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
- [x] I have updated the README.md (if applicable)
- [x] I have added tests & descriptions to my models (and macros if applicable)
- [ ] I have added an entry to CHANGELOG.md

PR #171, opened 2022-07-11
Fix - Make `values` a list of list, allowing to load fields with commas

### LINKED ISSUE #170: Issue with loading data from the graph when columns have a `,` (including the JSON string)
Debugging the integration tests failure made me realize that there is some issue/limitation around how we load data for connectors that don't support the `values('a', 'b', 'c')` approach. 

Fixing it will require some decent refactor.

Currently, we are saving the "rows" of data to load as a list of string and do the following processing:
- for Postgres/Snowflake, we load it "as-is" by wrapping the string with parenthesis: `('a', 'b', 'c')`
- for the other ones, we split reach row into individual column, using `value.split(',')`. While this works when no field has a `,` in it, it is actually failing if a field has one (like a JSON string with multiple keys).

I think that we should actually save the different rows as a list of lists rather than a list of strings but this requires a change to all the `get_` macros as well as `select_from_values`.

I will raise another issue about this..

_Originally posted by @b-per in https://github.com/dbt-labs/dbt-project-evaluator/issues/169#issuecomment-1180119224_

b-per:
This is a:
- [ ] new functionality

## Link to Issue 

Closes #170

## Description & motivation

Changing the logic of the different ingestion macros and `select_from_values` to make values a list of lists. We used to have it as a list of strings and split by `,` for BigQuery and Redshift but it doesn't work if there is a `,` in a field, like `meta` or if we want to load a JSON string in a column.

## Checklist
    - [x] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [x] Snowflake

### DIFF
--- a/macros/unpack/get_exposures.sql
+++ b/macros/unpack/get_exposures.sql
@@ -13,17 +13,17 @@
 
           {%- set values_line = 
             [
-              "'" ~ node.unique_id ~ "'",
-              "'" ~ node.name ~ "'",
-              "'" ~ node.resource_type ~ "'",
-              "'" ~ node.original_file_path ~ "'",
+              wrap_string_with_quotes(node.unique_id),
+              wrap_string_with_quotes(node.name),
+              wrap_string_with_quotes(node.resource_type),
+              wrap_string_with_quotes(node.original_file_path),
               "cast(" ~ dbt_project_evaluator.is_not_empty_string(node.description) | trim ~ " as boolean)",
-              "'" ~ node.type ~ "'",
-              "'" ~ node.maturity~ "'",
-              "'" ~ node.package_name ~ "'",
-              "'" ~ node.url ~ "'",
-              "'" ~ node.owner.name ~ "'",
-              "'" ~ node.owner.email~ "'"
+              wrap_string_with_quotes(node.type),
+              wrap_string_with_quotes(node.maturity),
+              wrap_string_with_quotes(node.package_name),
+              wrap_string_with_quotes(node.url),
+              wrap_string_with_quotes(node.owner.name),
+              wrap_string_with_quotes(node.owner.email)
             ]
           %}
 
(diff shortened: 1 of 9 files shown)


macros/unpack/get_exposures.sql
```diff
@@ -11,21 +11,22 @@
...
-            '{{ node.maturity}}',
-            '{{ node.package_name }}',
-            '{{ node.url }}',
-            '{{ node.owner.name }}',
-            '{{ node.owner.email }}'
-
-          {% endset -%}
+          {%- set values_line = 
+            [
+              "'" ~ node.unique_id ~ "'",
```
dave-connors-3, line 16: is there any way for us to make this a little less repetitive? i recently learned about `adapter.quote` -- not sure it even works here but could be cool

macros/unpack/get_metrics.sql
```diff
@@ -32,7 +19,26 @@
             {%- else -%}
                 ''
             {% endif -%}
-        {%- endset -%}
+          {% endset %}
+
+          {%- set values_line =
```
dave-connors-3, line 24: any reason to split these set blocks apart?

dave-connors-3 (approved):
🦞 -- this looks good, a few clarifying comments but I'm with you on the approach. You cannot merge until you review #132 😃

macros/unpack/get_exposures.sql
b-per: I think that `adapter.quote` is targeted at quoting database object names (e.g. surrounding the db name with `` for BQ, etc...), which is different from how to quote strings.

I get your point though and will create a macro for this use case to make it clearer than using the complex pattern everywhere.

b-per: I have pushed a commit with a new macro

macros/unpack/get_metrics.sql
b-per: Actually I can't get them together. I don't think I can use `if`s and `for`s within the variable definition itself.

PR #178, opened 2022-07-22
Fix/multi ref tests

### LINKED ISSUE #172: Allow relation tests in other directories
### Describe the feature
Relationship tests between layers / directories currently fail the `fct_test_directories` test - these tests are often going to be comparing models that exist in other directories so should allow these comparisons. 

I'm not aware of any. 

### Additional context
Not AFAIK

### Who will this benefit?
Anyone using multi layer projects with relationship tests. 

I could do if needed

dave-connors-3:
This is a:
- [ ] new functionality

## Link to Issue 
Closes #172 

## Description & motivation
Some of the package models that summarize test-model relationships did not take into account tests that may have relationships to multiple models, like a relationships test might have. 

Changes include:

- `int_model_test_summary` -- add some window functions and additional aggregation to ensure we don't double count the number of tests in the project
- `fct_missing_primary_key_tests` -- exclude some now irrelevant columns from the test summary model
- `fct_test_directories` -- only evaluate the directory logic for the model the test is applied to

image

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake


b-per:
> If we think we should adjust the logic to only count the model with the test directly applied in the yml config, we'll need a slight logic change here!

I fee like I'd count it only in the main/first model

PR #182, opened 2022-08-03
loop through all model_types for coverage models

### LINKED ISSUE #181: Documentation and test coverage for custom types
### Describe the feature

Currently, the `fct_documentation_coverage`  and `fct_test_coverage` models require users to have a 'marts' model type/folder in their project. Otherwise, the associated test will return an error with a division by zero.

CTE in the model where 'mart' is hard-coded.

```
conversion as (
    select
        resource_id,
        case when is_described then 1.0 else 0 end as is_described_model,
        case when model_type = 'marts' then 1.0 else NULL end as is_marts_model,
        case when is_described and model_type = 'marts' then 1.0 else 0 end as is_described_marts_model

    from models
),
```
I would like to be able to override the baked-in value for fact models and customize is through the `dbt_project.yml` file, just as in case of  tests.

My first though is to keep it in a `{{ var("mart_alias", "marts") }}`, which can be driven via project config, but still has it's default value.

### Additional context
This is a universal issue. It can also be the case that you (dbt Labs) want to enforce people using the same conventions, but still would be great to customize.

### Who will this benefit?
It would benefit those who want to tailor the evaluator to fit their folder structure. Personally, I'm not able to see these coverage percentages without having a folder named 'mart'

Absolutely! Although I might need a hand in finding out the best way we can tackle this.

graciegoheen:
This is a:
- [ ] new functionality

## Link to Issue 

Closes #181

## Description & motivation

Updated `fct_documentation_coverage` and `fct_test_coverage` to return the `_coverage_pct` for _each_ `model_type` listed in the `model_types` variable in your dbt_project.yml. 

Screen Shot 2022-08-03 at 3 58 33 PM

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,10 +438,11 @@ in the `tests` directory of you project.
 Specifically, this models measures:
 1. `test_coverage_pct`: the percentage of your models that have minimum 1 test applied.
 2. `test_to_model_ratio`: the ratio of the number of tests in your dbt project to the number of models in your dbt project
-3. `marts_test_coverage_pct`: the percentage of your marts models that have minimum 1 test applied.
+3. `< model_type >_test_coverage_pct`: the percentage of each of your model types that have minimum 1 test applied.
 
 This model will raise a `warn` error on a `dbt build` or `dbt test` if the `test_coverage_pct` is less than 100%.
-You can set your own threshold by overriding the `test_coverage_target` variable. [See overriding variables section.](#overriding-variables)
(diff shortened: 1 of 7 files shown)


b-per (approved):
Thanks Grace! I fixed the issue with the integration tests on databricks but it all lgtm now! 🥐

dave-connors-3 (approved):
🦞

PR #183, opened 2022-08-08
fix multi ref tests - alt approach

### LINKED ISSUE #172: Allow relation tests in other directories

dave-connors-3:
This is a:
- [ ] new functionality

## Link to Issue 
Closes #172 

## Description & motivation

Based on the feedback from the team in Slack, and in the prior PR, I thought this would be a cleaner approach. This creates a new boolean dimension for whether the test-model relationship is the primary relationship based on the test naming convention. Whichever model appears earlier in the test name would be considered the primary, and should map to the model in the user's yml file that has the test applied to it. There may be another simpler way to do this with the unpack macros, so keen to get some eyes on this!

This makes it such that the edits to the fct models only need to be a simple filter for the primary test/model relationship, and the logic remains relatively unchanged. 

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/macros/unpack/get_relationships.sql
+++ b/macros/unpack/get_relationships.sql
@@ -5,7 +5,6 @@
 {%- macro default__get_relationships(node_type) -%}
 
     {%- if execute -%}
-
         {%- if node_type == 'nodes' %}
             {% set nodes_list = graph.nodes.values() %}   
         {%- elif node_type == 'exposures' -%}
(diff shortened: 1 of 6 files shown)


macros/unpack/get_relationships.sql
```diff
@@ -38,7 +38,8 @@
                     {%- set values_line = 
                         [
                             "cast('" ~ node.unique_id ~ "' as " ~ dbt_utils.type_string() ~ ")",
-                            "cast('" ~ parent ~ "' as " ~ dbt_utils.type_string() ~ ")"
+                            "cast('" ~ parent ~ "' as " ~ dbt_utils.type_string() ~ ")",
+                            "" ~ loop.last ~ ""
```
b-per, line 42: Nice!

graciegoheen (approved):
Nice work on this! Looks super clean

PR #188, opened 2022-08-11
🩹 Fix `loop_vars` when there is no variable

### LINKED ISSUE #187: Need to update how we create "empty tables" across warehouses
### Describe the bug

Currently, in `loop_vars.sql` (and maybe other files, I need to check), we are handling the case where there is no data by creating an empty table that way.

```
select '' as var_name, null as var_value
where var_value is not null
```

This syntax works in Snowflake but doesn't on Postgres and BigQuery.

For BQ, I also tried:
```
select '' as var_name, null as var_value
where true = false
```
and got `Query without FROM clause cannot have a WHERE clause at [2:1]
`

I'll do some checks but the following might work across warehouses:
```
select '' as var_name, null as var_value
limit 0
```

b-per:
This is a:
- [ ] new functionality

## Link to Issue 

Closes #187

## Description & motivation

I didn't add a test as I couldn't see anything that we might really want in `integration_tests` but I have tested that it worked across the main DW even if there is the variable list is empty.

### DIFF
--- a/macros/loop_vars.sql
+++ b/macros/loop_vars.sql
@@ -22,7 +22,7 @@
 {%- if sql_query -%}
 {{ sql_query | join('union all') }}
 {%- else -%}
-select '' as var_name, null as var_value
-where var_value is not null
+select '' as var_name, '' as var_value
+limit 0
 {%- endif -%}
 {% endmacro %}


dave-connors-3 (approved):
🦞

Issue #193, opened 2022-08-17
Rules models all fail since upgrading to 0.1.2 (Postgres)


qmoret:
### Steps to reproduce
Upgrade to 0.1.2
Run with `dbt build --select package:dbt_project_evaluator --vars 'evaluator: true'`
I have also tried adding an exception seed with an example but no luck.

### Expected results
Rules models should run without error (actual behavior with 0.1.2)

### Actual results
All rules model fail with the same Database error, example below

### Screenshots and log output
Example : 
```
16:15:35  Database Error in model fct_model_naming_conventions (models/marts/structure/fct_model_naming_conventions.sql)
16:15:35    invalid input syntax for type integer: "fct_model_naming_conventions"
16:15:35    LINE 8: where fct_name = 'fct_model_naming_conventions'
```

### System information
packages:
  - package: dbt-labs/dbt_utils
    version: [">=0.8.0", "<0.9.0"]
  - package: dbt-labs/dbt_project_evaluator
    version: 0.1.2

**Which database are you using dbt with?**
- [x] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
Core:
  - installed: 1.2.0
  - latest:    1.2.0 - Up to date!

Plugins:
  - redshift: 1.2.0 - Up to date!
  - postgres: 1.2.0 - Up to date!
```

graciegoheen:
Thanks for catching this bug. I think this is an issue with the column types when the seed file is empty. I will dig into this further. 

In the meantime, when you added an exception seed with an example - can you confirm that you deactivated the original seed:

```
seeds:
  dbt_project_evaluator:
    dbt_project_evaluator_exceptions:
      +enabled: false
```
  
And ran both the package models and the seed file?

```
dbt build --select package:dbt_project_evaluator dbt_project_evaluator_exceptions 
```

qmoret:
Oh I hadn't run the seed file. But I get a similar issue, on the seed run this time : 
```
07:49:16  Database Error in seed dbt_project_evaluator_exceptions (seeds/dbt_project_evaluator_exceptions.csv)
07:49:16    invalid input syntax for type integer: "fct_multiple_sources_joined"
07:49:16    LINE 3:           ('fct_multiple_sources_joined','child','stg_%_unio...
```

b-per:
[...]

PR #195, opened 2022-08-18
Add new tests

b-per:
This is a:
- [ ] new functionality

## Description & motivation

Modify the CI process by adding a new project to handle use cases that we can't handle with the first one:
- no exposure
- no metric
- no overwrite of the default seed
- (we can add more later)

I have also updated the CI steps to run the second project after the first one

This is expected to fail now because of #193 but if the issue is the same as the one linked we should merge the code and then make sure that the fix for the issue above passes the CI tests.


b-per:
The error is now the same as the reported one, so I think it would be OK to merge this (currently failing) PR:
```
12:40:26  Database Error in model fct_source_directories (models/marts/structure/fct_source_directories.sql)
12:40:26    invalid input syntax for integer: "fct_source_directories"
12:40:26    LINE 8: where fct_name = 'fct_source_directories
```

integration_tests_2/README.md
```diff
@@ -0,0 +1,3 @@
+## Test dbt Project
+
+The models within this folder are used for testing use cases that can't be checked as part of the other `integration_tests` project. No tests are defined as the main purpose is to ensure that all models run properly in specific situations.
```
graciegoheen, line 3: Can we define what the edge cases are that this is capturing?

- no metrics
- no exposures
- no exceptions seed

dave-connors-3 (approved):
🦞

PR #200, opened 2022-09-09
update utils version and update cross db macros

### LINKED ISSUE #198: Loosen dependency requirements in `packages.yml`
### Describe the feature
Providing a range of supported package versions for `dbt_utils` would allow for `dbt_project_evaluator` to be installed alongside other packages much more easily. Currently, `dbt_project_evaluator` is [pinned to the `0.8.6` patch version of `dbt_utils`](https://github.com/dbt-labs/dbt-project-evaluator/blob/8269c67c0fe13320e0cafd7d5c79c272621e546b/packages.yml#L2-L3), which heavily constrains the dependency graph of compatible package versions when installed in conjunction with other packages.

The alternative is to only install `dbt_project_evaluator` alongside packages that are compatible with the pinned patch versions. 

### Additional context
N/A

### Who will this benefit?
Users who are attempting to update their dbt packages on a regular cadence in order to adopt the latest features and bugfixes.

I haven't contributed before. I would be happy to help, but I would need a hand getting started.

dave-connors-3:
This is a:
- [ ] new functionality

## Link to Issue 

Closes #198 

## Description & motivation

This PR loosens the package dependency on `dbt_utils` -- right now, we pin this package to `dbt_utils 0.8.6` -- this is causing some cross compatibility issues with other packages that now require/allow dbt utils v0.9.0 and above. 

After updating the `packages.yml` file, and installing utils 0.9.1, there were a number of warnings triggered because of our use of cross db macros in `dbt_utils` now available in dbt core. I updated the macros that are now available in core to remove the warnings. There are still a few warnings triggered:

- `array_append` and `array_concat` have not yet been migrated to core
- currently the `union_relation` macros in utils `0.9.1` still contains a cross db macro (string_literal) pinned to the package version, so that will continue to throw a log warning until all the utils utility macros also use the dbt-core versions. 

### Edit Sep 12th!

After talking with @joellabes, in order to release this PR to prevent the warnings related to the migration of the array related macros to core, I have moved those macros into this package in this PR. This allows us to remove dbt_utils entirely from the cross db macros process, and will make updating this package easier in the future when those macros are moved into dbt core, and we release a 1.3-friendly version of this package. 

image

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -16,6 +16,11 @@ clean-targets:         # directories to be removed by `dbt clean`
   - "target"
   - "dbt_packages"
 
+
+dispatch:
+  - macro_namespace: dbt
+    search_order: ['dbt_project_evaluator', 'dbt']
+
 models:
   dbt_project_evaluator:
     marts:
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -2,7 +2,7 @@ name: 'dbt_project_evaluator'
 version: '1.0.0'
 config-version: 2
 
-require-dbt-version: [">=1.1.0", "<2.0.0"]
+require-dbt-version: [">=1.2.0", "<2.0.0"]
 
 model-paths: ["models"]
 analysis-paths: ["analysis"]
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -16,10 +16,6 @@ clean-targets:         # directories to be removed by `dbt clean`
   - "target"
   - "dbt_packages"
 
-dispatch:
-  - macro_namespace: dbt
-    search_order: ['dbt_project_evaluator', 'dbt']
-
 models:
   dbt_project_evaluator:
     marts:
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -17,8 +17,6 @@ clean-targets:         # directories to be removed by `dbt clean`
   - "dbt_packages"
 
 dispatch:
-  - macro_namespace: dbt_utils
-    search_order: ['dbt_project_evaluator', 'spark_utils', 'dbt_utils']
   - macro_namespace: dbt
     search_order: ['dbt_project_evaluator', 'dbt']
 
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -19,6 +19,8 @@ clean-targets:         # directories to be removed by `dbt clean`
 dispatch:
   - macro_namespace: dbt_utils
     search_order: ['dbt_project_evaluator', 'spark_utils', 'dbt_utils']
+  - macro_namespace: dbt
+    search_order: ['dbt_project_evaluator', 'dbt']
 
 models:
   dbt_project_evaluator:
--- a/integration_tests/dbt_project.yml
+++ b/integration_tests/dbt_project.yml
@@ -25,8 +25,6 @@ clean-targets:         # directories to be removed by `dbt clean`
   - "dbt_packages"
 
 dispatch:
-  - macro_namespace: dbt_utils
-    search_order: ['dbt_project_evaluator', 'spark_utils', 'dbt_utils']
   - macro_namespace: dbt
     search_order: ['dbt_project_evaluator', 'dbt']
 
--- a/integration_tests/dbt_project.yml
+++ b/integration_tests/dbt_project.yml
@@ -24,10 +24,6 @@ clean-targets:         # directories to be removed by `dbt clean`
   - "target"
   - "dbt_packages"
 
-dispatch:
-  - macro_namespace: dbt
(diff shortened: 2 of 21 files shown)


packages.yml
```diff
@@ -1,5 +1,5 @@
 packages:
   - package: dbt-labs/dbt_utils
-    version: 0.8.6
+    version: [">=0.8.6", "<2.0.0"]
   - package: dbt-labs/spark_utils
```
dave-connors-3, line 4: @b-per -- actually pretty sure we can remove the dependency on spark utils in this PR -- does that sound right to you?

dbt_project.yml
```diff
@@ -17,8 +17,8 @@ clean-targets:         # directories to be removed by `dbt clean`
   - "dbt_packages"
 
 dispatch:
-  - macro_namespace: dbt_utils
-    search_order: ['dbt_project_evaluator', 'spark_utils', 'dbt_utils']
+  - macro_namespace: dbt
+    search_order: ['dbt_project_evaluator', 'dbt']
```
joellabes, line 21: If you dispatch the macros below to your own package (as discussed) then you won't need this at all - as you identified, it feels wrong because it is!

integration_tests/dbt_project.yml
```diff
@@ -25,8 +25,8 @@ clean-targets:         # directories to be removed by `dbt clean`
   - "dbt_packages"
 
 dispatch:
-  - macro_namespace: dbt_utils
-    search_order: ['dbt_project_evaluator', 'spark_utils', 'dbt_utils']
+  - macro_namespace: dbt
+    search_order: ['dbt_project_evaluator', 'dbt']
```
joellabes, line 29: Ditto here (and everywhere else)

macros/cross_db_shim/array_append.sql
```diff
@@ -0,0 +1,23 @@
+{# Moving this macro into this package to release a dbt-core 1.2 friendly release that does not rely on old dbt_utils versions #}
+{# These macro will be deprecated when we release a dbt-core 1.3 friendly release of this pacakge, which will include these macros#}
+
+
+{% macro array_append(array, new_element) -%}
+  {{ return(adapter.dispatch('array_append', 'dbt')(array, new_element)) }}
```
joellabes, line 6: These should dispatch to your package's name

macros/recursive_dag.sql
```diff
@@ -59,7 +59,7 @@ all_relationships (
         directory_path as child_directory_path,
         file_name as child_file_name,
         0 as distance,
-        {{ dbt_utils.array_construct(['resource_name']) }} as path,
+        {{ array_construct(['resource_name']) }} as path,
```
joellabes, line 62: And then these should be `dbt_project_evaluator.array_construct()` for now and swap out to `dbt.array_construct()` after 1.3

joellabes (changes requested):
@dave-connors-3 it's all the stuff I slacked you, but in the actual place for code review!

dbt_project.yml
dave-connors-3: roger that captain!

macros/recursive_dag.sql
dave-connors-3: unqualified name should also work here, correct?

dave-connors-3: i.e.`{{ array_append() }}`

packages.yml
dave-connors-3: looks like we were just using this for the spark shims, which we have now moved into this package until dbt 1.3 is released! we are safe to remove it

macros/recursive_dag.sql
dave-connors-3: ( i updated to follow your suggestino, but both seem to work and just confirming my understanding)

joellabes: Uhhhhh in my experience, macros inside of a package have always had to be qualified. Were you installing this package into another project and then calling it, or just testing from inside of itself?

dave-connors-3: testing from within the nested `integrations_test` subdirectory

b-per (approved):
If all the comments from @joellabes have been solved it's good to go for me.

As discussed, we might want to do some extra work moving a couple of the spark macros to other repos but this looks critical enough to go as-is for now.

joellabes (approved):
Recommend opening an issue on Core to get the listagg stuff upstreamed too (the deadline for 1.3 is COMING UP QUICK) but otherwise lgtm 🦖

PR #208, opened 2022-09-16
added steps for ci check to readme

graciegoheen:
This is a:
- [x] new functionality

## Description & motivation

Outlined how to use this package for a CI check in the README

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/README.md
+++ b/README.md
@@ -921,7 +921,7 @@ By default the tests in this package are configured with "warn" severity, we can
 1. Create an environment variable to define the appropriate severity for each environment. In dbt Cloud, for example, we can easily create an environment variable `DBT_PROJECT_EVALUATOR_SEVERITY` that is set to "error" for the Continuous Integration environment and "warn" for all other environments:
 ![Creating DBT_PROJECT_EVALUATOR_SEVERITY environment variable in dbt Cloud](https://user-images.githubusercontent.com/53586774/190683057-cf38d8dd-de70-457c-b65b-3532dc8f5ea1.png)
 
-Note: It is also possible to use an evironment variable for dbt Core, but the actual implementation will depend on how dbt is orchestrated. 
+Note: It is also possible to use an environment variable for dbt Core, but the actual implementation will depend on how dbt is orchestrated. 
 
 2. Update you project.yml file to override the default severity for all tests in this package:
 ```yml
--- a/README.md
+++ b/README.md
@@ -25,7 +25,9 @@ Check [dbt Hub](https://hub.getdbt.com/dbt-labs/dbt_project_evaluator/latest/) f
 ### Additional setup for Databricks/Spark
 
 In your `dbt_project.yml`, add the following config:
-```yaml
+```yml
+# dbt_project.yml
+
 dispatch:
   - macro_namespace: dbt_utils
     search_order: ['dbt_project_evaluator', 'spark_utils', 'dbt_utils']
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -904,6 +906,34 @@ We then run both the seed and the package by executing the following command:
 dbt build --select package:dbt_project_evaluator dbt_project_evaluator_exceptions
 ```
 
+----
+## Running this package as a CI check
+
+Once you have addressed all current misalignments in your project (either by fixing them or configuring exceptions), you can use this package as a CI check to ensure code changes don't introduce new misalignments. The setup will vary based on whether you are using dbt CLoud or dbt core, but the general steps are as follows:
```
b-per, line 912:
```suggestion
Once you have addressed all current misalignments in your project (either by fixing them or configuring exceptions), you can use this package as a CI check to ensure code changes don't introduce new misalignments. The setup will vary based on whether you are using dbt Cloud or dbt Core, but the general steps are as follows:
```

README.md
```diff
@@ -904,6 +906,34 @@ We then run both the seed and the package by executing the following command:
...
+
+### 1. Override test severity with an environment variable
+
+By default the tests in this package are configured with "warn" severity, we can override that for our CI jobs with an environment variable:
+1. Create an environment variable to define the appropriate severity for each environment. In dbt Cloud, for example, we can easily create an environment variable `DBT_PROJECT_EVALUATOR_SEVERITY` that is set to "error" for the Continuous Integration environment and "warn" for all other environments:
```
b-per, line 917: Could we mention that it is also possible with env vars for dbt Core, but the actual implementation will depend on how dbt is run/orchestrated?

README.md
```diff
@@ -904,6 +906,34 @@ We then run both the seed and the package by executing the following command:
...
+![Creating DBT_PROJECT_EVALUATOR_SEVERITY environment variable in dbt Cloud](https://user-images.githubusercontent.com/53586774/190683057-cf38d8dd-de70-457c-b65b-3532dc8f5ea1.png)
+
+2. Update you project.yml file to override the default severity for all tests in this package:
+```
+tests:
+  dbt_project_evaluator:
+    +severity: "{{ env_var('DBT_PROJECT_EVALUATOR_SEVERITY', 'warn') }}"
+```
+
+Note: you could follow a similar process to disable the models in this package for your production environment
```
b-per, line 927: If we want to talk about deactivating tests I think we should then mention that it will be using `enabled`true/false  instead of `severity` warn/error.

README.md
```diff
@@ -904,6 +906,34 @@ We then run both the seed and the package by executing the following command:
...
+tests:
+  dbt_project_evaluator:
+    +severity: "{{ env_var('DBT_PROJECT_EVALUATOR_SEVERITY', 'warn') }}"
+```
+
+Note: you could follow a similar process to disable the models in this package for your production environment 
+
+### 2. Run this package for each pull request
+
+Now, you can run this package as a step of your CI job/pipeline. In dbt Cloud, for example, you could simply add a command to your CI job:
```
b-per, line 931: Could we add the command in `code` as well so people can copy/paste it into their own config?

b-per (changes requested):
Thanks Grace! Couple of comments from my side

README.md
graciegoheen: Yep!

b-per (approved):

b-per:
The failing CI steps look odd. I am wondering if we didn't have 2 CI checks running at the same time that conflicted with each other. I'll trigger it again.

PR #233, opened 2022-11-15
added parent_model_type to fct_model_fanout

### LINKED ISSUE #230: include `parent_model_type` in `fct_model_fanout` to enable easier exception writing based on model type
### Describe the feature
 It would be great to add a new column to `fct_model_fanout` that contains the model_type (mart/staging/etc) of the parent model.

We are working around this by upping the `models_fanout_threshold` to 10. However, this means that we could miss fanout of fct/dim models which is something we would like to prevent

### Additional context
When looking at a relatively small dbt instance, fanout can be an unimportant issue especially when it is happening for stg or int models. However, we still believe that fanout is important to identify for fct/dim models. With this it would be easier to write an exception for the tests on the `fct_model_fanout` that exclude stg/int models.

### Who will this benefit?
This would benefit AE teams in ensuring following standard practices.

graciegoheen:
This is a:
- [x] new functionality

## Link to Issue 

Closes #230 

## Description & motivation

Added field `parent_model_type` to `fct_model_fanout` to allow users to write exceptions based off of the parent's model type. 

Screen Shot 2022-11-15 at 4 17 41 PM

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
    - [ ] Databricks
    - [ ] DuckDB

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,6 +495,7 @@ Tests are assertions you make about your models and other resources in your dbt 
 
 <details>
 <summary><b>How to Remediate</b></summary>
+
 Apply a [uniqueness test](https://docs.getdbt.com/reference/resource-properties/tests#unique) and a [not null test](https://docs.getdbt.com/reference/resource-properties/tests#not_null) to the column that represents the grain of your model in its schema entry. For models that are unique across a combination of columns, we recommend adding a surrogate key column to your model, then applying these tests to that new model. See the [`surrogate_key`](https://github.com/dbt-labs/dbt-utils#surrogate_key-source) macro from dbt_utils for more info!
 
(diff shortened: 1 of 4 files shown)


dave-connors-3 (approved):
🦞

b-per:
Just wondering, there is no test implication with this change? And if not, should there be either a change or a new test?

dave-connors-3:
@b-per good callout, we should update the tests too. I think it passed because we're hardcoding the columns to compare in the equality test! 

should be a quick fix!

graciegoheen:
Updated the integration test @b-per

b-per (approved):
🥖

PR #235, opened 2022-11-18
allow for unique combo of columns

### LINKED ISSUE #192: Include dbt_utils.unique_combination_of_columns and expression for missing_primary_key_tests
### Describe the feature
Currently the missing_primary_tests only checks for the uniqueness test of a single column.

It would be great to include these options:
https://docs.getdbt.com/faqs/tests/uniqueness-two-columns

### LINKED ISSUE #221: Uniqueness Tests
### Describe the feature
Extend or make configurable uniqueness test so it recognises unique_combination_of_columns from dbt_utils

Making a single unique column on the table

### Additional context
No

### Who will this benefit?
People who use dat-vault in particular for satellite models. Additionally fact tables in dimensional marts

### LINKED ISSUE #231: Support other primary key definitioms
### Describe the feature
Currently, [fct_missing_primary_key_tests](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/models/marts/tests/fct_missing_primary_key_tests.sql) checks if a model has a primary key based on having a `not_null` and `unique` test. These tests could both potentially apply to a non-primary key, or a different test could be used to verify a primary key. Saliently for me, I am using the [primary key test in DBT Constraints from Snowflake Labs](https://github.com/Snowflake-Labs/dbt_constraints/tree/0.5.2/#installation) and not unique/not_null.

It would be great if I could pass a test name as a project variable that matches my primary key test name to work with the primary key test coverage in this package.

[Additional field for other primary key definitions could be parsed here](https://github.com/dbt-labs/dbt-project-evaluator/blob/ad8440b50d201526fc94b9484784f71b608f07f0/models/marts/core/int_all_graph_resources.sql)

Disabling/overriding models in this package with my custom model. This is very inconvenient and may be incompatible with future changes.

### Who will this benefit?
This will certainly benefit users of DBT constraints.

Yes, happy to add this feature if it is supported by the package authors.

dave-connors-3:
This is a:
- [ ] new functionality

## Link to Issue 
Closes #231 
Closes #221 
Closes #192 

## Description & motivation
This PR both

- expands the default acceptable methodology for primary key tests to include `dbt_utils.unique_combination_of_columns` (per our docs site [here](https://docs.getdbt.com/faqs/tests/uniqueness-two-columns))
- gives users the ability to supply their own method(s) of primary key tests to the package to check with the variable `primary_key_test_macros` 

This new variable allows a user to supply a list. Each entry in the list is either a single test name or a list of test names that together comprise primary key tests. To illustrate, the default value set in `dbt_project.yml` is:
```yml
primary_key_test_macros: [["macro.dbt.test_unique", "macro.dbt.test_not_null"], "macro.dbt_utils.test_unique_combination_of_columns"]
```
This denotes that for each model, it must have either a unique _**and**_ not null test on a single column or have the unique combination of columns macro to be considered a pass for this check. 

Users like @ttusing who use the primary key test from `dbt_constraints` would just need to override the default value of this variable to include `macro.dbt_constraints.test_primary_key`. 

image

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB


dave-connors-3:
once we finalize the approach, i'll add to the readme to capture this new approach!

graciegoheen:
Thoughts on changing the variable from 
```
primary_key_test_macros: [["macro.dbt.test_unique", "macro.dbt.test_not_null"], "macro.dbt_utils.test_unique_combination_of_columns"]
```

to 
```
primary_key_test_macros: [["dbt.test_unique", "dbt.test_not_null"], "dbt_utils.test_unique_combination_of_columns"]
```

and adding the "macro." piece on as part of the string matching?

b-per:
Super nice Dave! 🚀 

My question would be more about the default value for `primary_key_test_macros`. Should we use `var('primary_key_test_macros', '[["macro.dbt.test_unique", "macro.dbt.test_not_null"]]'` in the code to default the value to `unique` and `not_null` if no overwrite is provided?

Also, being a new function, we would up the minor number when it is released and we should have some details under `README.md` that might include a gotcha. For example, if someone puts
```
primary_key_test_macros: ["macro.dbt.test_unique", "macro.dbt.test_not_null"]
```
it will consider that having `unique` OR `not_null` makes the column a PK, so it is important to double the `[` with
```
primary_key_test_macros: [["macro.dbt.test_unique", "macro.dbt.test_not_null"]]
```

dave-connors-3:
@b-per @graciegoheen on board with both your suggestions! I like the idea of only having to remember `project_name.test_name` instead of the whole qualified name, and definitely think it's worthwhile to adjust the README. 

@b-per -- no issue from me including the default value in-line, but hoping you could explain how that is functionally different  from including the default value in the `dbt_project.yml`?

b-per:
> hoping you could explain how that is functionally different from including the default value in the dbt_project.yml

I would rather not clutter the `dbt_project.yml` if `unique` + `not_null` work for 90% of the people. For all those users, no var will be showing up in their project config file.

graciegoheen:
I actually prefer keeping all defaults in the `dbt_project.yml` file, so that it's easy to quickly understand what is customizable...

README.md
```diff
@@ -495,15 +500,17 @@ Tests are assertions you make about your models and other resources in your dbt
...
+
+Apply a [uniqueness test](https://docs.getdbt.com/reference/resource-properties/tests#unique) and a [not null test](https://docs.getdbt.com/reference/resource-properties/tests#not_null) to the column that represents the grain of your model in its schema entry. For models that are unique across a combination of columns, we recommend adding a surrogate key column to your model, then applying these tests to that new model. See the [`surrogate_key`](https://github.com/dbt-labs/dbt-utils#surrogate_key-source) macro from dbt_utils for more info! Alternatively, you can use the [`dbt_utils.unique_combination_of_columns`](<https://github.com/dbt-labs/dbt-utils#unique_combination_of_columns-source>) test from `dbt_utils`. Check out the [overriding variables section](#overriding-variables) to read more about configuring other primary key tests for your project!
```
graciegoheen, line 504: Can we call out here that while you can use `dbt_utils.unique_combination_of_columns`, you risk all of those columns being NULL which is not technically a primary key?

README.md
```diff
@@ -888,21 +895,42 @@ models:
...
 
 <details>
-<summary><b>Coverage Variables</b></summary>
+<summary><b>Testing and Documentation Variables</b></summary>
 
 | variable    | description | default     |
 | ----------- | ----------- | ----------- |
 | `test_coverage_pct` | the minimum acceptable test coverage percentage | 100% |
 | `documentation_coverage_pct` | the minimum acceptable documentation coverage percentage | 100% |
+| `primary_key_test_macros` | the set(s) of test macro names used to test primary key macros | [["dbt.test_unique", "dbt.test_not_null"], "dbt_utils.test_unique_combination_of_columns"] |
```
graciegoheen, line 904: I'm confused by this "the set(s) of test macro names used to test primary key macros" - can we update it to something a bit clearer. Maybe something like "the set(s) of dbt tests used to check validity of a primary key"

README.md
```diff
@@ -888,21 +895,42 @@ models:
...
+
+Therefore, the `dbt.test_unique` and `dbt.test_not_null` are grouped together in a sub-list to ensure they are evaluated together. 
+
+*While it's not explicitly tested in this package, we strongly encourage adding a `not_null` test on each of the columns listed in the `dbt_utils.unique_combination_of_columns` tests.*
+
 
 ```yml
 # dbt_project.yml
 # set your test and doc coverage to 75% instead
+# use the dbt_constraints.test_primary_key test for fct_missing_primary_key_tests.sql
```
graciegoheen, line 926:
```suggestion
# use the dbt_constraints.test_primary_key test to check for validity of your primary keys
```

README.md
dave-connors-3: I call that out [down here](https://github.com/dbt-labs/dbt-project-evaluator/pull/235/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R920)

Is that the wrong spot for it?

graciegoheen: That's good with me!! Just wanted to make sure it was somewhere

models/marts/core/int_all_graph_resources.sql
```diff
@@ -1,4 +1,17 @@
 -- one row for each resource in the graph
+
+{# flatten the sets of permissable primary key test sets to one level for later iteration #}
+{%- set test_macro_list = [] %}
+{%- for test_set in var('primary_key_test_macros') -%}
+    {%- if test_set is iterable and (test_set is not string and test_set is not mapping) -%}
```
graciegoheen, line 6: Rather than needing to have different behavior for when the test set is a string vs. a list, I wonder if we should just require that each input for the primary_key_test_macros is a list. So the default would be:

`[["dbt.test_unique", "dbt.test_not_null"], ["dbt_utils.test_unique_combination_of_columns"]]`

and the override example would be:

`primary_key_test_macros: [["dbt_constraints.test_primary_key"]]`

Maybe that's worse because it requires you to remember that each set needs [] but would make this piece of code less complex.

models/marts/tests/intermediate/int_model_test_summary.sql
```diff
@@ -13,7 +13,19 @@ count_column_tests as (
     select 
         relationships.direct_parent_id, 
         all_graph_resources.column_name,
-        count(distinct case when all_graph_resources.is_unique_test or all_graph_resources.is_not_null_test then relationships.resource_id else null end) primary_key_tests_count,
+        {%- for test_set in var('primary_key_test_macros') %}
+            {%- set outer_loop = loop -%}
+            {%- if test_set is iterable and (test_set is not string and test_set is not mapping) %}
```
graciegoheen, line 18: Same comment here... we're adding complexity by allowing such a wide range of acceptable inputs for the primary_key_test_macros variable.

models/marts/tests/intermediate/int_model_test_summary.sql
```diff
@@ -27,7 +39,17 @@ agg_test_relationships as (
 
     select 
         direct_parent_id, 
-        sum(case when primary_key_tests_count = 2 then 1 else 0 end) >= 1 as is_primary_key_tested,
+        sum(case 
+                when (
+                    {%- for test_set in var('primary_key_test_macros') %}
+                        {%- set compare_value = test_set | length if (test_set is not string and test_set is not mapping) else 1 %}
```
graciegoheen, line 45: And here as well

graciegoheen (commented):
I don't think this is blocking, but left some comments on the primary_key_test_macros variable. Curious what you think.

README.md
```diff
@@ -892,21 +898,39 @@ models:
...
+**Usage notes for `primary_key_test_macros:`**
+
+The `primary_key_test_macros` variable determines how the `fct_missing_primary_key_tests` ([source](models/marts/tests/fct_missing_primary_key_tests.sql)) model evaluates whether the models in your project are properly tested for their grain. This variable is a list and each entry **must be a list of test names in `project_name.test_macro_name` format**.
+
+For each entry in the parent list, the logic in `int_model_test_summary` will evaluate whether each model has all of the tests in that entry applied. If a model meets the criteria of any of the entries in the parent list, it will be considered a pass. The default behavior for this package will check for whether each model has either:
+
+1. __Both__ the `not_null` and `unique` tests applied to a single column OR
+2. The `dbt_utils.unique_combination_of_columns` applied to the model.
+
+Therefore, the `dbt.test_unique` and `dbt.test_not_null` are grouped together in a sub-list to ensure they are evaluated together.
```
graciegoheen, line 918: Now that we're requiring a list of lists, maybe this should be updated

README.md
```diff
@@ -892,21 +898,39 @@ models:
...
 ```yml
 # dbt_project.yml
 # set your test and doc coverage to 75% instead
+# use the dbt_constraints.test_primary_key test to check for validity of your primary keys
 
 vars:
   dbt_project_evaluator:
     documentation_coverage_target: 75
     test_coverage_target: 75
+    primary_key_test_macros: ["dbt_constraints.test_primary_key"]
```
graciegoheen, line 932: I believe this should now be `[["dbt_constraints.test_primary_key"]]`

README.md
dave-connors-3: updated this line very slightly, but if you have something else in mind let me know!

graciegoheen:
```suggestion
Therefore, each set of test(s) that define a primary key requirement must be grouped together in a sub-list to ensure they are evaluated together (e.g. [`dbt.test_unique`, `dbt.test_not_null`] ).
```

graciegoheen (approved):
L(ord) G(race) T(hinks) M(agnificent)!

PR #238, opened 2022-11-25
fix broken links

patkearns10:
This is a:
- [ ] new functionality

## Description & motivation

The context index links were leading to the index itself and not the section further down


b-per (approved):
🥐

Issue #189, opened 2022-08-12
raw table reference fact model


dave-connors-3:
### Describe the feature

Create a model that parses SQL using regex to identify raw table references in model code

### Who will this benefit?

All y'all

christineberger:
➕ To this one! I would love to see a table like the following:
| model | explicit_reference |
|-------|-------------------|
| lost_revenue | raw_tpch.tpch_sf1.orders |
| lost_revenue | raw_tpch.tpch_sf1.customer |
| lost_revenue | raw_tpch.tpch_sf1.lineitem |
| lost_revenue | raw_tpch.tpch_sf1.nation |
| UK_Lowcost_Brass_Suppliers | raw_tpch.tpch_sf1.partsupp |

Throwing some example code here which uses the `tpch_sf1` data sets, although the second references some other models
I have in my project. Feel free to ping me when this is being worked on if you need access to the code for testing! I should have a complete version then, but right now these models are a WIP for a Coalesce workshop so there's not a good version I can link here. 

I have two models which use explicit references:
`lost_revenue.sql`:
```sql
WITH

-- Only orders which tie back to a real customer record
ro as (
    select
        o.o_orderkey,
        c.c_custkey
    from raw_tpch.tpch_sf1.orders o
    inner join raw_tpch.tpch_sf1.customer c
        on o.o_custkey = c.c_custkey
),

rl as (
    SELECT c.c_custkey
    ,c.c_name
    ,sum(revenue) AS revenue_lost
    ,c_acctbal
    ,n.n_name 
    ,c_address 
    ,c_phone 
    ,c_comment
    FROM ro left join (SELECT l.l_orderkey AS ORDER_ID, o.o_custkey AS CUSTOMER_ID, SUM(l.l_extendedprice * (1 - l.l_discount)) AS REVENUE
        FROM raw_tpch.tpch_sf1.lineitem l LEFT JOIN raw_tpch.tpch_sf1.orders o ON l.l_orderkey = o.o_orderkey
        WHERE o.o_orderdate >= '1994-01-01' AND o.o_orderdate < '1994-04-01' AND l.l_returnflag = 'R' GROUP BY o.o_custkey, l.l_orderkey
    ) lo on lo.ORDER_ID = ro.o_orderkey and lo.CUSTOMER_ID = ro.c_custkey
[...]

PR #246, opened 2022-12-02
feature/add new model fct_hard_coded_references to capture all models that have hard coded references

### LINKED ISSUE #240: Handle/flag hard-coded references
### Describe the feature
I would love for this package to be able to handle (and flag?) hard-coded references (ie `raw.stripe.payment` vs. `{{ source ('stripe', 'payment') }`). I specifically noticed that 'fct_unused_sources` currently returns sources that are accessed only via hard-coded references instead of the `source` or `ref` macros. There may be other applications for this functionality within the package as well.

The minimum response to this issue would be to add a note to the `readme` regarding the results of the `fct_unused_sources` model to include this possibility. A value-add in that documentation could be a recommendation about how to discern whether the sources are truly unused vs. accessed through hard-coding. For instance, in my case I used the list generated by `fct_root_models` as a starting place when looking for hard-coded references.

### Additional context
This should apply to multiple databases.

### Who will this benefit?
This would benefit anyone with hard-coded references in their project, allowing them to systematically transition to appropriate alternatives (`source` and `ref` functions). 

Sure! And I'd be glad for some discussion/input around how to approach solving this.

graciegoheen:
This is a:
- [x] new functionality

## Link to Issue 

Closes #240 and #189

## Description & motivation

Create a new model `fct_hard_coded_references` that captures all models that have hard coded reference(s) in their raw sql

TO DO:
- add text to readme
- consider only returning the unique list of raw references, in alphabetic order
- should the list be limited if too big?
- add documentation to dag.yml

Screen Shot 2022-12-02 at 2 41 26 PM

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
    - [ ] Databricks
    - [ ] DuckDB


graciegoheen:
Hmm - I think it could make sense to load regexes from another package but this model vs. codegen use different regexes with some overlap. Might be worth brainstorming what exactly the regex macros should look like, since these two use cases are different. 

Is this blocking? Or do you want me to open up a new issue for this package & codgen but continue finalizing this PR?

joellabes:
If they're similar but not the same, trying to shoehorn them into the same behaviour is probably a bad idea. 

> should they be in `dbt-codegen` or should they be in their own package and be imported both here and in `dbt-codegen`

If you're going to do this, I'd rather its own package - codegen shouldn't have any direct relationship to project-evaluator. 

IMO you could continue with it duplicated, but if you have a third use case where this comes up then extracting it probably makes sense. Very unscientific heuristics here so feel free to disagree!

macros/find_all_raw_references.sql
```diff
@@ -0,0 +1,238 @@
...
+
+                    {%- do all_raw_references_list.append(raw_reference) -%}
+
+                {%- endfor -%}
+        
+        {%- endfor -%}
+
+    {% endif %}
+
+    # TO DO: consider returning in alphabetic order, and only unique values!
```
b-per, line 232: From version `1.2` dbt has a [`set`](https://docs.getdbt.com/reference/dbt-jinja-functions/set) command (the Python `set`) that can be used to remove duplicates from a list (among other things)

macros/find_all_raw_references.sql
```diff
@@ -0,0 +1,238 @@
...
+            'from_var_1':
+                '(?ix)
+
+                # first matching group
+                # from or join followed by at least 1 whitespace character
+                (from|join)\s+
+
+                # second matching group
+                # opening {{, 0 or more whitespace character(s), var, 0 or more whitespace character(s), an opening parenthesis, 0 or more whitespace character(s), 1 or 0 quotation mark
+                ({{\s*var\s*\(\s*[\'\"]?)
```
b-per, line 52: So beautiful ❤️

README.md
```diff
@@ -269,6 +270,33 @@ or if you want to use base_ models and keep stg_model_2 as is:
 <img width="500" alt="A refactored DAG showing two base models feeding into a staging model" src="https://user-images.githubusercontent.com/30663534/159602135-926f2823-3683-4cd5-be00-c04c312ed42d.png">
 </details>
 
+#### Raw References
```
b-per, line 273: Is it really raw reference? Or missing `source`/`ref`?

README.md
```diff
@@ -269,6 +270,33 @@ or if you want to use base_ models and keep stg_model_2 as is:
...
+`fct_raw_references` ([source](models/marts/dag/fct_raw_references.sql)) shows each instance where a model contains raw references.
+
+<details>
+<summary><b>Example</b></summary>
+
+blah blah
```
b-per, line 279: Agree 👍

b-per:
It looks great and super useful.
I added some comments but they are about minor things.

README.md
graciegoheen: Maybe hard-coded references?

README.md
```diff
@@ -164,6 +165,77 @@ After refactoring your downstream model to select from the staging layer, your D
...
+Always use the `ref` function when selecting from another model and the `source` function when selecting from raw data, rather than using the direct relation reference (e.g. `my_schema.my_table`). Direct relation references are determined via regex mapping [here](macros/find_all_hard_coded_references.sql). 
+
+The `ref` and `source` functions are part of what makes dbt so powerful! Using these functions allows dbt to infer dependencies (and check that you haven't created any circular dependencies), properly generate your DAG, and ensure that models are built in the correct order. This also ensures that your current model selects from upstream tables and views in the same environment that you're working in.
+
+</details>
+
+<details>
+<summary><b>Exceptions</b></summary>
+
+TO DO: Are there any exceptions anyone can think of?? I think there might be some packages that select from a variable...
```
b-per, line 207: Should we remove the TODO?

graciegoheen: yes, but this is an open question - can you think of any exceptions?

b-per:
Looks good to me overall. I added a comment about the TODO in the README and we just need to get databricks passing as well.

README.md
b-per: No, and if there is any, then people will just use the seed file I guess

macros/cross_db_shim/spark_shims.sql
```diff
@@ -0,0 +1,3 @@
+{% macro spark__escape_single_quotes(expression) -%}
```
graciegoheen, line 1: This should eventually be added to the spark adapter!

b-per (approved):
Thanks Grace! LGTM 🥐

integration_tests/seeds/dag/dag_seeds.yml
```diff
@@ -73,4 +73,13 @@ seeds:
...
-            - is_loop_independent
\ No newline at end of file
+            - is_loop_independent
+
+  - name: test_fct_hard_coded_references
+    tests:
+      - dbt_utils.equality:
+          name: equality_fct_hard_coded_references
+          compare_model: ref('fct_hard_coded_references')
+          compare_columns:
```
dave-connors-3, line 83: nit: since the compare columns are all the columns, we can just leave out this config

macros/cross_db_shim/spark_shims.sql
dave-connors-3: i'm honestly floored by the whitespace issue adding spaces

dave-connors-3 (approved):
great work! -- stray thought: is this in service of replacing our current `fct_root_models` check? I feel like the overlap here would be really close to 100%

PR #248, opened 2022-12-05
`source --> exposure` check

### LINKED ISSUE #247: check for exposures dependent on sources
### Describe the feature
during a client audit I found an exposure that had a `source` in the `depends_on` field -- would like to either incorporate sources into the existing exposure model or make that more clear in the docs if that is not already the case!

### Who will this benefit?
exposure users near and far

oh you bet

dave-connors-3:
This is a:
- [x] new functionality

## Link to Issue 

Closes #247 

## Description & motivation

This adds a new `structure` model to check for exposures dependent directly on sources. best practices indicate that all sources should be modeled in some way before being exposed to downstream tools

Alternative is to make the `fct_parent_model_materializations` model more multipurpose -- I chose to do this for a first cut because this check seems more spiritually aligned with a DAG check rather than a performance related check. 

Will add docs once we agree on the strategy here!

image

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB


integration_tests/models/marts/exposures.yml
```diff
@@ -13,4 +13,5 @@ exposures:
     
     depends_on:
       - ref('fct_model_6')
-      - ref('dim_model_7')
\ No newline at end of file
+      - ref('dim_model_7')
+      - source('source_1', 'table_1')
```
dave-connors-3, line 17: i do not see this newline in my local editor 🤔

b-per (commented):
Nice @dave-connors-3 
My only comment, for now, would be whether this looks a bit "too niche" or not.

Thinking about our different types of nodes, would it be worth modifying this one into a `fct_exposures_metrics_dependent_on_sources.sql`. We would most likely want `metrics` to not depend on sources as well, no?

README.md
```diff
@@ -849,7 +849,7 @@ The best practice to determine top candidates for changing materialization from
 
 #### Exposure Parents Materializations
 
-`fct_exposure_parents_materializations` ([source](models/marts/performance/fct_exposure_parents_materializations.sql)) shows each model that is a direct parent of an exposure and is *not* materialized as a table in the warehouse. 
+`fct_exposure_parents_materializations` ([source](models/marts/performance/fct_exposure_parents_materializations.sql)) shows each resource that is a direct parent of an exposure and is not a model that is materialized as a table in the warehouse.
```
graciegoheen, line 852: Aren't we also allowing metrics?

graciegoheen (approved):
nice !

PR #250, opened 2022-12-06
exclude custom test directory from check

### LINKED ISSUE #249: custom test creates issue in `fct_test_directories`
### Describe the bug

the logic in `fct_test_directories` treats custom tests  in your `tests/` directory the same as tests applied in model yml files. The goal of the check is to ensure that tests in yml are applied in the same subdirectory as the models they test, but for custom tests, this is not true! we should exclude these from the check

[link to slack](https://getdbt.slack.com/archives/C046EL9GCJU/p1670323923957289)

### Steps to reproduce

1. add a custom test to your `tests` directory
2. `dbt build -s +fct_test_directories`
3. see custom test in output

image

### Expected results

only tests in the models directory

### Actual results

all tests

### System information

- [x] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake

```
<output goes here>
```

dave-connors-3:
This is a:
- [ ] new functionality

## Link to Issue 
Closes #249 

## Description & motivation

This PR excludes tests not in the `models` directory from being checked by the `fct_test_directories` model. Custom tests are in the right place already!

image

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB

### DIFF
--- a/models/marts/core/int_all_graph_resources.sql
+++ b/models/marts/core/int_all_graph_resources.sql
@@ -44,6 +44,7 @@ joined as (
         unioned_with_calc.resource_type, 
         unioned_with_calc.file_path, 
         unioned_with_calc.directory_path,
+        unioned_with_calc.directory_path like 'models/%' as is_in_models_directory,
         unioned_with_calc.file_name,
         case 
             when unioned_with_calc.resource_type in ('test', 'source', 'metric', 'exposure', 'seed') then null
--- a/integration_tests/seeds/tests/test_fct_missing_primary_key_tests.csv
+++ b/integration_tests/seeds/tests/test_fct_missing_primary_key_tests.csv
@@ -7,6 +7,6 @@ int_model_5,FALSE,0
 report_1,FALSE,0
 report_2,FALSE,0
 report_3,FALSE,0
-stg_model_1,FALSE,1
+stg_model_1,FALSE,2
 stg_model_3,FALSE,0
 stg_model_5,FALSE,0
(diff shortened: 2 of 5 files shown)


models/marts/core/int_all_graph_resources.sql
```diff
@@ -53,6 +53,7 @@ joined as (
         unioned_with_calc.resource_type, 
         unioned_with_calc.file_path, 
         unioned_with_calc.directory_path,
+        unioned_with_calc.directory_path like 'models/%' as is_in_models_directory,
```
graciegoheen, line 56: Is there a way to get the `model-paths` from dbt_project.yml file instead of hard-coding?

graciegoheen: Nice! That's better than hard-coding, so I'll take it :)

macros/unpack/get_nodes.sql
```diff
@@ -25,7 +25,8 @@
                 "cast(" ~ dbt_project_evaluator.is_not_empty_string(node.description) | trim ~ " as boolean)",
                 "''" if not node.column_name else wrap_string_with_quotes(dbt.escape_single_quotes(node.column_name)),
                 wrap_string_with_quotes(node.meta | tojson),
-                wrap_string_with_quotes(node.depends_on.macros | tojson)
+                wrap_string_with_quotes(node.depends_on.macros | tojson),
+                wrap_string_with_quotes(node.test_metadata.name if node.test_metadata else ''),
```
graciegoheen, line 29: What if we just went ahead and calculated `is_generic_test` here instead of pulling in the `node.test_metadata.name` value?

dave-connors-3: good question! mostly comes down to whether we think the test name (or other metadata) will be valuable in other contexts

wasilaq (approved):
LGTM!

graciegoheen (approved):
nice !

Issue #241, opened 2022-12-01
Generalize model relationship patterns (without relying on prefixes)


deanna-minnick:
### Describe the feature
I recently noticed three models returned no results for a project I was reviewing, even though it's likely that they have some of these patterns in their project. Upon closer inspection I realized their model prefixes don't match the assumptions utilized in this package, so none of their dependencies were flagged in:
- `fct_staging_dependent_on_marts_or_intermediate`
- `fct_staging_dependent_on_staging`
- `fct_marts_or_intermediate_dependent_on_source`

In my happy dream world, the model relationship patterns would still be flagged somehow, even if a project's models aren't specifically prefixed by modeling layer. 

### Describe alternatives you've considered
The bare minimum would be an update to the `readme` that acknowledges the limitations of these three models (ie they'll only detect these model relationship patterns if you're utilizing `stg`, `int`, `fct`, `dim`, etc. prefixes).

### Additional context
This is project-specific (ie naming conventions), not database-specific.

### Who will this benefit?
This will make the package more usable for folks who don't use prefixes that correlate to modeling layers.

### Are you interested in contributing this feature?
Possibly! Definitely glad to propose a `readme` updated and may be able to contribute to a more substantial solution as time allows.

PR #264, opened 2022-12-20
clarify modeling layer assumptions in README

deanna-minnick:
This is a:
- [X] new functionality

## Link to Issue 
This addresses open issue #241.

## Description & motivation
This PR updates the README to clarify the assumption that resources are named by modeling layer (whether conforming exactly to dbt Labs' style guide or not) and explicitly acknowledges limitations in these checks if resources are named using another categorization method.

N/A

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB
- [X] I have updated the README.md (if applicable)

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Each test warning indicates the presence of a type of misalignment. To troublesh
 3. Either fix the issue(s) or [customize](#customization) the package to exclude them
 
 #### Caveat Regarding Modeling Layer Assumptions
-This package assumes that you are using standard modeling layers (ie staging, intermediate, and marts) and that those layers are reflected in both the folder structure and naming prefixes used in your project. If you utilize these modeling layers but have prefixes other than `stg_`, `int_`, `fct_`, and `dim_`, you may override which prefixes the package should look for (see [Overriding Variables](#overriding-variables)). If your project doesn't use these modeling layers, you may wish to ignore or disable package models that don't apply in your case (ie [Downstream Models Dependent on Source](#downstream-models-dependent-on-source)).
+This package assumes that you are using standard modeling layers (ie staging, intermediate, and marts) and that those layers are reflected in both the folder structure and naming prefixes used in your project. If you utilize these modeling layers but have prefixes other than `stg_`, `int_`, `fct_`, and `dim_`, you may override which prefixes the package should look for (see Naming Convention Variables in [Overriding Variables](#overriding-variables)). If your project doesn't use these modeling layers, you may wish to ignore or disable package models that don't apply in your case (ie [Downstream Models Dependent on Source](#downstream-models-dependent-on-source)).
 
 ----
 ## Package Documentation
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ After refactoring your downstream model to select from the staging layer, your D
 `fct_marts_or_intermediate_dependent_on_source` ([source](models/marts/dag/fct_marts_or_intermediate_dependent_on_source.sql)) shows each downstream model (`marts` or `intermediate`)
 that depends directly on a source node.
 
-If you get no results or unexpected results, see [Modeling Layer Assumptions](#caveat-regarding-modeling-layer-assumptions) for more context.
+If this model returns no results, either your project doesn't contain any instances of this issue (hooray!) or you may need to update your [naming convention variables](#naming_convention_variables).
 
 <details>
 <summary><b>Example</b></summary>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -137,6 +137,8 @@ After refactoring your downstream model to select from the staging layer, your D
 `fct_marts_or_intermediate_dependent_on_source` ([source](models/marts/dag/fct_marts_or_intermediate_dependent_on_source.sql)) shows each downstream model (`marts` or `intermediate`)
 that depends directly on a source node.
 
+Note: if your naming conventions don't **correlate to modeling layer** (ie `stg`, `int`, `fct`, etc. or another custom naming pattern defined in the [Overriding Variables](#overriding-variables) section below), this check will return no results.
```
graciegoheen, line 140:
```suggestion
Note: if your naming conventions don't **correlate to relevant modeling layers** (i.e. `marts` and `intermediate`), this check will return no results.
```
I think this is currently hard-coded

README.md
```diff
@@ -50,6 +50,9 @@ Each test warning indicates the presence of a type of misalignment. To troublesh
 2. Query the associated model to find the specific instances of the issue within your project
 3. Either fix the issue(s) or [customize](#customization) the package to exclude them
 
+#### Caveat Regarding Modeling Layer Assumptions
+This package assumes that you are using standard modeling layers (ie staging, intermediate, and marts) and that those layers are reflected in both the folder structure and naming prefixes used in your project. If you utilize these modeling layers but have prefixes other than `stg_`, `int_`, `fct_`, and `dim_`, you may override which prefixes the package should look for (see [Overriding Variables](#overriding-variables)). If your project doesn't use these modeling layers, you may wish to ignore or disable package models that don't apply in your case (ie [Downstream Models Dependent on Source](#downstream-models-dependent-on-source)).
```
graciegoheen, line 54: I feel like this is more of a default rather than an assumption? 

And I would avoid listing out the default prefixes here since they're already captured in the Overriding Variables section.

deanna-minnick: It's definitely a default, and I would argue it's an assumption since there isn't really any way to configure or use these five models if you don't use modeling layers of some kind. 

Regarding removing the default prefixes, I see your point from an updatability perspective, but I also think these values are unlikely to change (and certainly not frequently). It feels like a trade-off in usability (ie requiring a second click if you got unexpected results from your model), and I'd be inclined to leave it as is. How strongly do you feel about this?

graciegoheen: I'm going to add a comment with suggested changes so we can see how it would read

graciegoheen: Just want to avoid repeating code since we already have a very similar paragraph:

> The model_types, <model_type>_folder_name, and <model_type>_prefixes variables allow the package to check if models in the different layers are in the correct folders and have a correct prefix in their name. The default model types are the ones we recommend in our [dbt Labs Style Guide](https://github.com/dbt-labs/corp/blob/main/dbt_style_guide.md). If your model types are different, you can update the model_types variable and create new variables for <model_type>_folder_name and/or <model_type>_prefixes.

graciegoheen: How would you feel about just linking to this section instead for each of the relevant models?

README.md
```diff
@@ -137,6 +140,8 @@ After refactoring your downstream model to select from the staging layer, your D
 `fct_marts_or_intermediate_dependent_on_source` ([source](models/marts/dag/fct_marts_or_intermediate_dependent_on_source.sql)) shows each downstream model (`marts` or `intermediate`)
 that depends directly on a source node.
 
+If you get no results or unexpected results, see [Modeling Layer Assumptions](#caveat-regarding-modeling-layer-assumptions) for more context.
```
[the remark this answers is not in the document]
deanna-minnick: Truth! Appreciate the feedback - updates will be committed momentarily!

README.md
```diff
@@ -50,6 +50,9 @@ Each test warning indicates the presence of a type of misalignment. To troublesh
 2. Query the associated model to find the specific instances of the issue within your project
 3. Either fix the issue(s) or [customize](#customization) the package to exclude them
 
+#### Caveat Regarding Modeling Layer Assumptions
+This package assumes that you are using standard modeling layers (ie staging, intermediate, and marts) and that those layers are reflected in both the folder structure and naming prefixes used in your project. If you utilize these modeling layers but have prefixes other than `stg_`, `int_`, `fct_`, and `dim_`, you may override which prefixes the package should look for (see Naming Convention Variables in [Overriding Variables](#overriding-variables)). If your project doesn't use these modeling layers, you may wish to ignore or disable package models that don't apply in your case (ie [Downstream Models Dependent on Source](#downstream-models-dependent-on-source)).
```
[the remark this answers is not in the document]
deanna-minnick: This definitely feels like an improvement - thanks! Preference for keeping this or just getting rid of it since it's so similar to the Naming Convention Variable section?

README.md
deanna-minnick: Oh, good call - totally down for this! I've updated the links accordingly (not 100% sure the anchor link is working as expected, since that section isn't a header).

I'm wondering if we just ditch the other paragraph I had created then? Unless we want to call this out explicitly early in the readme.

PR #277, opened 2023-01-05
check if seed exists in warehouse before filtering

### LINKED ISSUE #225: Avoid `dbt run` failing if the seed has not been loaded
### Describe the feature

A few people reported issues when they run the project without the seed being executed. A `dbt seed` or `dbt build` works (and this is what we have described in the README) but might not be super easy to catch at first.

Instead of having a "default seed" that needs to be run, I was wondering if we could remove it and use [get_relation](https://docs.getdbt.com/reference/dbt-jinja-functions/adapter#get_relation) to check if the seed exists in the warehouse or not. And if it doesn't exist we just don't filter out any of the `fct` tables.

Any opinion from anyone?

graciegoheen:
This is a:
- [ ] new functionality

## Link to Issue 

Closes #225 

## Description & motivation

Our readme suggests executing a `dbt build` of this package, which works properly. If users instead execute a `dbt run` on this package, it will fail as the blank seed necessary for the filtering exceptions functionality does not yet exist in their warehouse. 

I added a check to the filter_exceptions macro to first check if the seed exists in the warehouse, and if it doesn't, don't do any filtering.

Note: We still have to have the blank seed file in our project as the default, because otherwise this check will return a compilation error as the `dbt_project_evaluator_exceptions` would not be a node in the project..

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
    - [ ] Databricks
    - [ ] DuckDB

### DIFF
--- a/macros/filter_exceptions.sql
+++ b/macros/filter_exceptions.sql
@@ -4,9 +4,13 @@
 
 {% macro default__filter_exceptions(model_ref) %}
 
-    {% set seed_exists = load_relation(ref('dbt_project_evaluator_exceptions')) is not none %}
+    {% if execute %}
+    {% set is_custom_seed = graph.nodes.values() | 
+            selectattr('resource_type', 'equalto', 'seed') | 
+            selectattr('name', 'equalto', 'dbt_project_evaluator_exceptions') | 
+            selectattr('package_name', 'equalto', 'dbt_project_evaluator') is none %}
 
-    {% if seed_exists %}
+    {% if is_custom_seed %}
 
         {% set query_filters %}
         select
@@ -23,6 +27,8 @@
             {% endfor %}
         {% endif %}
     
+    {% endif %}
+
     {% endif %}
 
 {% endmacro %}
--- a/macros/filter_exceptions.sql
+++ b/macros/filter_exceptions.sql
@@ -20,7 +20,7 @@
 
     {% if not is_default_seed %}
     
-        {% if execute and flags.WHICH not in ['compile'] %}
+        {% if flags.WHICH not in ['compile'] %}
             where 1 = 1
             {% for row_filter in run_query(query_filters) %}
                 and {{ row_filter[0] }} not like '{{ row_filter[1] }}'
--- a/macros/filter_exceptions.sql
+++ b/macros/filter_exceptions.sql
@@ -1,22 +1,28 @@
 {% macro filter_exceptions(model_ref) -%}
-  {{ return(adapter.dispatch('filter_exceptions', 'dbt_project_evaluator')(model_ref)) }}
+    {{ return(adapter.dispatch('filter_exceptions', 'dbt_project_evaluator')(model_ref)) }}
 {%- endmacro %}
 
 {% macro default__filter_exceptions(model_ref) %}
 
-{% set query_filters %}
-select
-    column_name,
-    id_to_exclude
-from {{ ref('dbt_project_evaluator_exceptions') }}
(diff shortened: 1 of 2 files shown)


dave-connors-3 (approved):
LGTM

b-per:
I thought it was not possible but it looks like you got it working 😄 . On my side, I was focusing on the fact that it is currently not possible to put a conditional logic on the existence of a given model name in a dbt project when you are actually checking whether a model has been materialized or not, rather than if it exists in the project 🧠 

I am then thinking that the downside 😉 is that if people defined their own seeds but didn't do a `dbt build` or a `dbt seed` on it, the project will report incorrect exceptions instead of just failing; but I think that we have the same problem today if people materialize the original seed and forget to materialize their own new one.

Also, could we add a test somehow about it in our integration tests?

graciegoheen:
Yes, ideally we could check for the existence of a model name in a dbt project.. but it's just not possible today. I think this is at least an improvement for those that aren't using a custom exceptions seed. And for those that are, it'll be more intuitive to actually build the seed if they've created a custom one. 

Hmmm.. what kind of integration test should I add? We already have the integration_tests_2 folder which checks that `dbt build` works without a custom seed. Should we update circle-ci to also try executing a `dbt run`? @b-per

macros/filter_exceptions.sql
```diff
@@ -1,22 +1,28 @@
...
 {% macro default__filter_exceptions(model_ref) %}
 
-{% set query_filters %}
-select
-    column_name,
-    id_to_exclude
-from {{ ref('dbt_project_evaluator_exceptions') }}
-where fct_name = '{{ model_ref.name }}'
-{% endset %}
+    {% set seed_exists = load_relation(ref('dbt_project_evaluator_exceptions')) is not none %}
```
dave-connors-3, line 7:
```suggestion
    {% set seed_exists = adapter.load_relation(ref('dbt_project_evaluator_exceptions')) is not none %}
```

maybe?

graciegoheen: check which package this relation is from?

graciegoheen:
New logic: Now it will check which package the seed belongs to and if it belongs to the `dbt_project_evaluator` package it won't do any filtering

dave-connors-3 (approved):
that rips

macros/filter_exceptions.sql
```diff
@@ -1,22 +1,34 @@
...
+    {% if execute %}
+    {% set is_default_seed = 'dbt_project_evaluator' in graph.nodes.values() | 
+        selectattr('resource_type', 'equalto', 'seed') | 
+        selectattr('name', 'equalto', 'dbt_project_evaluator_exceptions') | 
+        map(attribute = 'package_name') | list %}
 
-{% endmacro %}
+    {% if not is_default_seed %}
+    
+        {% if execute and flags.WHICH not in ['compile'] %}
```
dave-connors-3, line 23: do you need this second if execute?

b-per (approved):
Thanks Grace! 

Once Dave's comment has been looked at it looks like this solution works well both when people define and don't define exceptions!

macros/filter_exceptions.sql
graciegoheen: nope!

graciegoheen:
Screen Shot 2023-01-23 at 10 44 09 AM

Note: `dbt run` completes successfully when you don't override the default seed file (`integration_tests_2`)

Issue #278, opened 2023-01-06
Allow single quotes in metric labels to be handled correctly


JakeC-MyT:
### Describe the bug
Currently, if a metric label contains single quotes e.g. `total users who answered 'yes' to X`, the `stg_metrics` model fails as the `get_metrics` macro doesn't escape these single quotes

### Steps to reproduce
Create a metric with single quotes in the label and run `dbt build --select package:dbt_project_evaluator`

### Expected results
Single quotes in metric labels to be escaped, similar to how they are in the filters list, where `dbt.escape_single_quotes` is utilised.

### Actual results
Single quotes are not escaped, leading to CTE build failing during `stg_metrics` build

### Screenshots and log output
`12:14:05  Database Error in model stg_metrics (models/staging/graph/stg_metrics.sql)
12:14:05    001003 (42000): SQL compilation error:
12:14:05    syntax error line 7 at position 20 unexpected 'values'.
12:14:05    syntax error line 11 at position 4 unexpected '('.
12:14:05    syntax error line 11 at position 5 unexpected 'True'.
12:14:05    syntax error line 11 at position 21 unexpected ','.
12:14:05    compiled Code at target/run/dbt_project_evaluator/models/staging/graph/stg_metrics.sql`

### System information
**The contents of your `packages.yml` file:**
`packages:
  - package: dbt-labs/dbt_utils
    version: 1.0.0

  - package: entechlog/dbt_snow_mask
    version: 0.2.3

  - package: dbt-labs/metrics
    version: 1.3.2

  - package: dbt-labs/dbt_project_evaluator
    version: 0.3.0`

**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [x] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
Core:
  - installed: 1.3.0
  - latest:    1.3.2 - Update available!

  Your version of dbt-core is out of date!
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation

Plugins:
  - bigquery:   1.3.0 - Up to date!
  - databricks: 1.3.0 - Update available!
  - firebolt:   1.2.0 - Not compatible!
  - postgres:   1.3.0 - Update available!
  - redshift:   1.3.0 - Up to date!
  - snowflake:  1.3.0 - Up to date!
  - spark:      1.3.0 - Up to date!

  At least one plugin is out of date or incompatible with dbt-core.
[...]

PR #279, opened 2023-01-06
Escape single quotes in metric labels in get_metrics()

JakeC-MyT:
This is a:
- [ ] new functionality

## Link to Issue 
https://github.com/dbt-labs/dbt-project-evaluator/issues/278

## Description & motivation
As discussed in the issue, currently the `get_metrics` macro can't handle single quotes in the label of a metric, causing a failure when building `stg_metrics`. This code adds a statement which sets `label` to be the single-quote escaped version of the label, and then passes that to the `wrap_string_with_quotes` function instead.

image

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
    - [ ] Databricks
    - [ ] DuckDB

### DIFF
--- a/integration_tests/models/marts/metrics.yml
+++ b/integration_tests/models/marts/metrics.yml
@@ -2,7 +2,7 @@ version: 2
 
 metrics:
   - name: new_customers
-    label: New Customers
+    label: New Customers marked 'paying'
     model: ref('dim_model_7')
     description: "The number of paid customers using the product"
 
--- a/integration_tests/models/marts/metrics.yml
+++ b/integration_tests/models/marts/metrics.yml
@@ -26,4 +26,3 @@ metrics:
 
     meta: 
       team: "Finance"
-      refresh_rate: "Bob's weekly run"
--- a/integration_tests/models/marts/metrics.yml
+++ b/integration_tests/models/marts/metrics.yml
@@ -26,3 +26,4 @@ metrics:
 
     meta: 
       team: "Finance"
+      refresh_rate: "Bob's weekly run"
--- a/macros/unpack/get_metrics.sql
+++ b/macros/unpack/get_metrics.sql
@@ -30,7 +30,7 @@
             "cast(" ~ dbt_project_evaluator.is_not_empty_string(node.description) | trim ~ " as boolean)",
             wrap_string_with_quotes(node.type),
             wrap_string_with_quotes(node.model.identifier),
-            wrap_string_with_quotes(dbt.escape_single_quotes(node.label))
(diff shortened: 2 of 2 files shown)


macros/unpack/get_metrics.sql
```diff
@@ -30,7 +34,7 @@
             "cast(" ~ dbt_project_evaluator.is_not_empty_string(node.description) | trim ~ " as boolean)",
             wrap_string_with_quotes(node.type),
             wrap_string_with_quotes(node.model.identifier),
-            wrap_string_with_quotes(node.label),
+            wrap_string_with_quotes(label),
```
dave-connors-3, line 37:
```suggestion
            wrap_string_with_quotes(dbt.escape_single_quotes(node.label))
```

dave-connors-3 (changes requested):
Thanks @JakeC-MyT so much for the issue and PR! This is a good straightforward fix. 

I added a suggestion to make this a one-line change rather than creating a new jinja variable in the macro.

Additionally, I would love to have this case covered in our integration tests! I think the simplest way to do that would be to edit the existing metric in `integration_tests/models/marts/metrics.yml` to include a single quote in the label, and ensure our CI checks pass!

macros/unpack/get_metrics.sql
JakeC-MyT: Thanks @dave-connors-3 - totally bizarre I'm sure I tried this one line fix the first time around and it didn't work, but running it again now it looks fine! 

Thank you for the advice on the integration test - sorry I'm not too familiar with them so wasn't sure how best to demonstrate it works, will add a new commit now :)

integration_tests/models/marts/metrics.yml
```diff
@@ -26,4 +26,4 @@ metrics:
 
     meta: 
       team: "Finance"
-      refresh_rate: "Bob's weekly run"
\ No newline at end of file
+      refresh_rate: "Bob's weekly run"
```
dave-connors-3, line 29: was this a net new addition? not sure why this is flagged as a change here

dave-connors-3 (approved):
This looks great! thanks so much for the contribution! will merge once we get one more approval

integration_tests/models/marts/metrics.yml
JakeC-MyT: Not net new no - seems to be showing as a deletion followed by replacement. Sorry, may have got my branch in a bit of a tizz - I can try and resolve if you need, but don't think it has any difference from what's in main. Let me know, and thanks for your support in contributing - very happy to be involved!

graciegoheen (approved):
Nice work!

PR #285, opened 2023-01-10
remove suffix instead of trimming prefix

### LINKED ISSUE #283: Naming Conventions - folders under "models" cannot have "_" in them due to the way model_type is extracted
### Describe the bug
Due to the way model_type is extracted [here](https://github.com/dbt-labs/dbt-project-evaluator/blob/80525c16a328c8067e220a5a55aff8d384be66eb/models/staging/variables/stg_naming_convention_prefixes.sql#L18), you cannot use a model prefix and folder name that has "_" in them. Configuring the `dbt_project_evaluator` vars with a model type with underscores in it (i.e "folder_with_underscores") and/or setting "folder_with_underscores_prefixes" variable doesn't work with a value like `['folder_with_underscores_']`.

### Steps to reproduce
Create a directory under the models directory, that its name has an underscore in it, like "folder_with_underscores" with a sample model in it. 

### Expected results
Models under the folder_with_underscores_ prefix do not show in fct_model_naming_conventions and do not fail is_empty_fct_model_naming_conventions_ test.

### Actual results
is_empty_fct_model_naming_conventions_ test fails and these models appear in fct_model_naming_conventions

### System information
```yaml
dbt_project_evaluator:
    model_types: ['staging', 'intermediate', 'marts', 'other', 'util', 'folder_with_underscores']
    folder_with_underscores_prefixes: ['folder_with_underscores_']
```

- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [X ] snowflake

```
Core:
  - installed: 1.3.1
  - latest:    1.3.2 - Update available!

  Your version of dbt-core is out of date!
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation

Plugins:
  - snowflake: 1.3.0 - Up to date!```


### Additional context
[Possible line of code that causes that](https://github.com/dbt-labs/dbt-project-evaluator/blob/80525c16a328c8067e220a5a55aff8d384be66eb/models/staging/variables/stg_naming_convention_prefixes.sql#L18)

### PR DESCRIPTION
**Pull request by dave-connors-3**
This is a:
- [ ] new functionality

## Link to Issue 
closes #283 

## Description & motivation

Allows for users to have model types that contain underscores. 

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB

### DIFF
--- a/integration_tests/dbt_project.yml
+++ b/integration_tests/dbt_project.yml
@@ -59,6 +59,8 @@ seeds:
 
 vars:
   # ensure integration tests run successfully when there are 0 of a given model type (extra)
-  model_types: ['staging', 'intermediate', 'marts', 'other', 'extra']
+  model_types: ['staging', 'intermediate', 'marts', 'other', 'extra', 'new_model_type']
   # dummy variable used for testing fct_hard_coded_references
   my_table_reference: 'grace_table'
+  new_model_type_folder_name: 'my_new_models'
+  new_model_type_prefixes: 'nwmdl_'
--- a/integration_tests/models/dbt_project_evaluator_schema_tests/graph.yml
+++ b/integration_tests/models/dbt_project_evaluator_schema_tests/graph.yml
@@ -64,6 +64,10 @@ models:
         tests:
           - unique
           - not_null
+      - name: model_type 
+        tests:
+          - accepted_values:
+              values: "{{ var('model_types') }}"
 
   - name: stg_naming_convention_prefixes
     description: "This table shows one record for each accepted model prefix for each model type (staging, intermediate, marts) from the provided variables"
@@ -73,3 +77,7 @@ models:
         tests:
           - unique
           - not_null
+      - name: model_type 
+        tests:
+          - accepted_values:
+              values: "{{ var('model_types') }}"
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: integration_tests/dbt_project.yml [hunk 1]
```diff
@@ -59,6 +59,8 @@ seeds:
...
 vars:
   # ensure integration tests run successfully when there are 0 of a given model type (extra)
-  model_types: ['staging', 'intermediate', 'marts', 'other', 'extra']
+  model_types: ['staging', 'intermediate', 'marts', 'other', 'extra', 'new_model_type']
   # dummy variable used for testing fct_hard_coded_references
-  my_table_reference: 'grace_table'
\ No newline at end of file
+  my_table_reference: 'grace_table'
+  new_model_type_folder_name: 'my_new_models'
+  new_model_type_prefixes: 'nwmdl_'
```
graciegoheen, line 66: Should we check that a prefix with _ works to? Like `nw_mods_`

dave-connors-3: I don't think we do anything to the value of these variables anywhere, we only do string manipulation on the name of the variable after we loop. 

One other thing we could do is refactor `loop_vars` to contain the suffix-appending logic, and then we might be able to remove the string parsing thing altogether

PR #294, opened 2023-01-31
add `distinct` back to relationships staging models

graciegoheen:
This is a:
- [x] bug fix PR with no breaking changes
- [ ] new functionality

## Description & motivation

The staging relationships models cannot have any duplicates otherwise it can lead to fanout in the downstream int_all_dag_relationships model. Adding a `distinct` back to the relevant staging relationships model

Note: Open to suggestions as to how to improve our integration tests here. There is already uniqueness checks for the `unique_id` of these models in integration_tests/models/dbt_project_evaluator_schema_tests/graph.yml

## Checklist
- [ ] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB
- [ ] I have updated the README.md (if applicable)
- [ ] I have added tests & descriptions to my models (and macros if applicable)

PR #296, opened 2023-02-01
🐛 Fix for duplicate PKs in stg relationships

### LINKED ISSUE #295: Duplicate entries in `stg_node_relationships` which also leads to later fanout
### Describe the bug

In the latest release (`0.4`) the model `stg_node_relationships` can contain duplicate records. This happens when the same table is `ref` or `source` multiple times in the same model.

Due to those duplicate records, it is possible that downstream models, like `int_all_dag_relationships` grow exponentially.

We have a test on the records being unique today but our mock project doesn't have a model with multiple `ref` to the same table.

### Steps to reproduce

Reproducing the issue in the integration tests:
image

image

### Are you interested in contributing the fix?

Yep

b-per:
This is a:
- [ ] new functionality

## Link to Issue 

Closes #295

## Description & motivation

I think that this superseeds #294 as well

I have added a test that was failing before this PR and is now passing

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [x] DuckDB

### DIFF
--- a/integration_tests/models/marts/exposures.yml
+++ b/integration_tests/models/marts/exposures.yml
@@ -14,4 +14,5 @@ exposures:
     depends_on:
       - ref('fct_model_6')
       - ref('dim_model_7')
+      - ref('dim_model_7')
       - source('source_1', 'table_1')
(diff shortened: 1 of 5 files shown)


dave-connors-3:
@b-per this looks good to me!

two thoughts:

1. should we add distinct to the other realtionships staging models as well? I don't think there's anything stopping you from doing the same thing with an exposure for example.
2. Can we add a dupe reference in an exposure and an extra source in a model in the integration tests to make sure we're covering our bases?

b-per:
I can't approve the PR as I am the original author but your changes LGTM Dave. OK to merge.

dave-connors-3 (approved):

PR #301, opened 2023-02-03
Allow override of `model_type` determination preference

### LINKED ISSUE #286: Allow preferential use of model folder when determining model type
### Describe the feature
I would like to be able to choose whether the model sub-folder or prefix takes preference when determining the model type.

I have overridden the `int_all_graph_resources` model so that the coalesce to determine the model type takes the model sub-folder in preference.  However, this isn't ideal as when we update the package, we will need to also remember to update the code here.

```sql
        , case
            when resource_type in ('test', 'source', 'metric', 'exposure', 'seed') then null
            /* Changed from normal to take the folder in preference to the prefix */
            else coalesce(model_type_folder, model_type_prefix, 'other')
        end as model_type
```

### Additional context
This is for a large project that has been migrated from a legacy system, so the models are structured slightly outside the norm. This means that sub-folders are used to determine the model type, more often than the model name (which is harder to change). It would be really useful to be able to specify that the model folder is more important in determining the model type than the model prefix.

### Who will this benefit?
Anyone who uses the same prefixes across multiple model types.

glsdown:
This is a:
- [x] new functionality

## Link to Issue 
Closes #286 

## Description & motivation
This allows users to override the default behaviour that uses the model prefix to determine the model type, in preference to the model folder.

I am unsure how to run this.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
    - [ ] Databricks
    - [ ] DuckDB

I am unsure how to add tests to this as it isn't a new test, but changing the default behaviour. I have verified locally that changing the variable changes the compiled SQL:

** Default behaviour**

```yml
# dbt_project.yml
vars:
  prefer_model_folder_type_to_prefix: false
```

```sql
# target/compiled/dbt_project_evaluator/models/marts/core/int_all_graph_resources.sql
...
        case 
            when resource_type in ('test', 'source', 'metric', 'exposure', 'seed') then null
            else 
                -- by default we will define the model type based on its prefix in the case prefix and folder types are different
                    coalesce(model_type_prefix, model_type_folder, 'other')
        end as model_type,
...
```

**Overridden behaviour**

```yml
# dbt_project.yml
vars:
  prefer_model_folder_type_to_prefix: true
```

```sql
# target/compiled/dbt_project_evaluator/models/marts/core/int_all_graph_resources.sql
...
        case 
            when resource_type in ('test', 'source', 'metric', 'exposure', 'seed') then null
            else 
                -- default behaviour overriden to define the model type based on its folder in the case prefix and folder types are different
                    coalesce(model_type_folder, model_type_prefix, 'other')
        end as model_type,
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1034,6 +1034,7 @@ vars:
 | `intermediate_prefixes` | the list of acceptable prefixes for your intermediate models | int_ |
 | `marts_prefixes` | the list of acceptable prefixes for your marts models | fct_, dim_ |
 | `other_prefixes` | the list of acceptable prefixes for your other models | rpt_ |
+| `prefer_model_folder_type_to_prefix` | override to prefer the use of the folder name to the prefix when determining model type | false |
 
 The `model_types`, `<model_type>_folder_name`, and `<model_type>_prefixes` variables allow the package to check if models in the different layers are in the correct folders and have a correct prefix in their name. The default model types are the ones we recommend in our [dbt Labs Style Guide](https://github.com/dbt-labs/corp/blob/main/dbt_style_guide.md). If your model types are different, you can update the `model_types` variable and create new variables for `<model_type>_folder_name` and/or `<model_type>_prefixes`.
 
@@ -1047,6 +1048,8 @@ vars:
     util_folder_name: 'util'
     util_prefixes: ['util_']
 ```
+
+By default, we use the `<model_type>_prefixes` to determine the `model_type` where the prefix and folder name types differ, however you can set the `prefer_model_folder_type_to_prefix` to `true` to change this behaviour to take the type determined by the folder name in preference.
 </details>
 
 <details>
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1034,6 +1034,7 @@ vars:
 | `intermediate_prefixes` | the list of acceptable prefixes for your intermediate models | int_ |
 | `marts_prefixes` | the list of acceptable prefixes for your marts models | fct_, dim_ |
 | `other_prefixes` | the list of acceptable prefixes for your other models | rpt_ |
+| `prefer_model_folder_type_to_prefix` | override to prefer the use of the folder name to the prefix when determining model type | false |
```
graciegoheen, line 1037: Hi! Could you update the description here to signal what the variable does, rather than what overriding it does?

dave-connors-3:
@glsdown are you still interested in contributing this change?

PR #303, opened 2023-02-06
Update int_all_graph_resources.sql

### LINKED ISSUE #302: fct_model_directories bug
### Describe the bug
fct_model_directories stores records for all the models that are not considered to be located appropriately in your porject according to the configuration of model types, folder names. If a model only belongs to a directory called Y_X' and another directory called 'X' is present and defined as a model type folder name in the dbt_project yml file then the dbt project evaluator believes it should appear under the directory called 'X' therefore producing a warn in the is_empty_fct_model_directoris test.

### Steps to reproduce
In a dbt project add version 0.4.0 of the dbt project evaluator to your dbt packages yml file and run `dbt deps`

In your dbt project yml file configure variables for the project evaluator as below

```
vars:
dbt_project_evaluator:  
    model_types: ['X', 'Y_X']
    X_folder_name: 'X'
    X_prefixes: ['X_']
    Y_X_folder_name: 'Y_X'
    Y_X_prefixes: ['Y_']
```

Create two directories 'X' and 'Y_X' and under the 'Y_X' folder create a model called 'Y_X_model.sql'

run `dbt build -s package:dbt_project_evaluator` 

### Expected results
As the 'Y_X_model.sql' is placed under the correct directory the 'is_empty_fct_model_directories' test should pass

### Actual results
The test presents a warn with a count of 1. The 1 record is obviously the 'Y_X_model.sql'.

image

I believe the bug comes from a combination of lines 62 to 66 ,

`case 
            when unioned_with_calc.resource_type in ('test', 'source', 'metric', 'exposure', 'seed') then null
            when {{ dbt.position('naming_convention_folders.folder_name_value','unioned_with_calc.directory_path') }} = 0 then null
            else naming_convention_folders.model_type 
        end as model_type_folder,`

line 115',

`row_number() over (partition by resource_id order by position_folder desc) as folder_name_rank`

and line 123  all in 'int_all_graph_resources.sql'

`where folder_name_rank = 1`

As 'X' appears in the folder name 'Y_X', two records appear with non_null value in the model_type_folder field. 
And 'X' appears later in the directory path so will given the folder_name_rank = 1 which is the criteria for the final selection.

I believe this can be solved by changing the jinja on line 64 to

`{{ dbt.position("concat('/' + 'naming_convention_folders.folder_name_value' + '/', 'unioned_with_calc.directory_path') }}`

Therefore when looking for '/X/' in the directory 'models/Y_X/Y_X_model.sql' the record will return a null value for that record.

### System information

- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [x ] snowflake

Using dbt cloud so dbt v1.3.0-latest

### Additional context
In my actual project the directories causing problems are called 'sats' and 'eff_sats' and there isn't a better naming convention I can use.

DVDylanRoe:
This is a:
- [x ] bug fix PR with no breaking changes
- [ ] new functionality

## Link to Issue 
Closes #302 

## Description & motivation
Before the script was identifying a model's type by finding the latest position a model type's folder name appears in the model's directory path. Now the folder name has been concatenated with a '/' before meaning the search is for a key directory not the just the key word.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB

### DIFF
--- a/models/marts/core/int_all_graph_resources.sql
+++ b/models/marts/core/int_all_graph_resources.sql
@@ -61,10 +61,10 @@ joined as (
         end as model_type_prefix,
         case 
             when unioned_with_calc.resource_type in ('test', 'source', 'metric', 'exposure', 'seed') then null
-            when {{ dbt.position("concat('/',naming_convention_folders.folder_name_value,'/')",'unioned_with_calc.directory_path') }} = 0 then null
+            when {{ dbt.position(dbt.concat(["'" + get_directory_pattern() + "'",'naming_convention_folders.folder_name_value',"'" + get_directory_pattern() + "'"]),'unioned_with_calc.directory_path') }} = 0 then null
             else naming_convention_folders.model_type 
         end as model_type_folder,
-        {{ dbt.position("concat('/',naming_convention_folders.folder_name_value,'/')",'unioned_with_calc.directory_path') }} as position_folder,  
+        {{ dbt.position(dbt.concat(["'" + get_directory_pattern() + "'",'naming_convention_folders.folder_name_value',"'" + get_directory_pattern() + "'"]),'unioned_with_calc.directory_path') }} as position_folder,  
         nullif(unioned_with_calc.column_name, '') as column_name,
         {% for test in test_macro_list %}
         unioned_with_calc.macro_dependencies like '%macro.{{ test }}%' and unioned_with_calc.resource_type = 'test' as is_{{ test.split('.')[1] }},  
--- a/models/marts/core/int_all_graph_resources.sql
+++ b/models/marts/core/int_all_graph_resources.sql
@@ -61,10 +61,10 @@ joined as (
         end as model_type_prefix,
         case 
             when unioned_with_calc.resource_type in ('test', 'source', 'metric', 'exposure', 'seed') then null
-            when {{ dbt.position('naming_convention_folders.folder_name_value','unioned_with_calc.directory_path') }} = 0 then null
+            when {{ dbt.position("concat('/',naming_convention_folders.folder_name_value,'/')",'unioned_with_calc.directory_path') }} = 0 then null
             else naming_convention_folders.model_type 
         end as model_type_folder,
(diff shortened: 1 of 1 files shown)


models/marts/core/int_all_graph_resources.sql
```diff
@@ -61,10 +61,10 @@ joined as (
         end as model_type_prefix,
         case 
             when unioned_with_calc.resource_type in ('test', 'source', 'metric', 'exposure', 'seed') then null
-            when {{ dbt.position('naming_convention_folders.folder_name_value','unioned_with_calc.directory_path') }} = 0 then null
+            when {{ dbt.position("concat('/',naming_convention_folders.folder_name_value,'/')",'unioned_with_calc.directory_path') }} = 0 then null
```
b-per, line 64: `concat()` doesn't work with more than 2 arguments on Redshift, hence why the failing CI check. 
You could most likely use [`dbt.concat`](https://docs.getdbt.com/reference/dbt-jinja-functions/cross-database-macros#concat) instead.

b-per, line 64: We also need to make it work for Windows users who have their files delimited with `\`.
There is [already a macro](https://github.com/dbt-labs/dbt-project-evaluator/blob/8a4c7e484a9ba81733c293d72ffeb447536a772a/macros/get_directory_pattern.sql#L2-L10) that you should be able to use that provides the file delimiter based on the OS being used.

b-per (changes requested):
Thanks for the PR!

I just left a couple of comments before we can get it merged.

DVDylanRoe:
@b-per replaced concat with dbt.concat and introduced get_directory_pattern() as requested. Has some issues as get_directory_pattern() wasn't returning the delimiter as a string once the code was compiled so the only work around I found is using "'" + get_directory_pattern() + "'".

b-per:
Thanks @DVDylanRoe and sorry for the delay. We are going to release a new version of the repo soon and we will take this change in account.

b-per:
Moving it to #317

PR #314, opened 2023-03-10
add checks for resource yml files

### LINKED ISSUE #312: add check to rename schema files to include a leading `_` and resource name
### Describe the feature

Per some internal conversation, the community team want to standardize the way folks name their resource yml files. 

Our recommended best practice is using a leading underscore, and including the resource type that is described within the yml file in the filename. For example, for a yml file with model configs, valid names for that file would be:

- _[resource_type].yml
- _[directory name]__[resource_type].yml

The underscore is for discoverability, the resource name for clarity. 

Do nothing, run wild, be free

### Who will this benefit?

`yml` writers of all kinds

giddy up

### LINKED ISSUE #313: ensure resource.yml files only contain one resource
### Describe the feature

There should be a model to enforce the best practice of not mixing resource types in one yml file. I.e. source definitons and model configs should live in separate `_models.yml` and `_sources.yml` files

### Additional context
Is this feature database-specific? Which database(s) is/are relevant? Please include any other relevant context here.

### Who will this benefit?
What kind of use case will this feature be useful for? Please be specific and provide examples, this will help us prioritize properly.

dave-connors-3:
This is a:
- [x] new functionality

## Link to Issue 

Closes #313 
Close #312 

## Description & motivation

We should add checks to ensure naming conventions for our resource yml files. 

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB


models/marts/structure/fct_resource_yml_naming_conventions.sql
```diff
@@ -0,0 +1,38 @@
...
+            "'s.yml'"
+        ]) }} as valid_filename_option_2
+        
+    
+    from all_graph_resources
+
+)
+
+select * from check_naming_convention
+where not(has_leading_underscore and has_plural_resource_name)
```
b-per, line 38: We are checking that there is an underscore at the beginning and a "s" at the end but we are actually not checking if the file is actually `valid_filename_option_1` or `valid_filename_option_2`, are we?

models/marts/core/int_all_graph_resources.sql
```diff
@@ -55,6 +58,19 @@ joined as (
         unioned_with_calc.directory_path,
         unioned_with_calc.is_generic_test,
         unioned_with_calc.file_name,
+        unioned_with_calc.patch_path,
+        -- use file_name when it's a yml only resource
```
b-per, line 62: The comment should be `file_path`, no?

models/marts/structure/fct_resource_yml_naming_conventions.sql
dave-connors-3: yeah, i built the booleans before building in the suggested renames -- makes more sense to filter where the name is not one of the valid ones!

Issue #234, opened 2022-11-17
Evaluate specific folder/file(s) to make remediation easier


jschintz-windriver:
### Describe the feature
When a project is old and littered with issues, provide a path for incremental improvements, focusing on keeping new code clean.  
2 options:
- Allow users to choose which folders/sub-folders are evaluated.
- Only evaluate the modified files and related.

### Describe alternatives you've considered
Is this a custom use case that the repository just needs forked?
Just start a new project from scratch?

### Who will this benefit?
Devs frequently inherit old code that never considered dbt best practices.
That legacy code is just too much to fix at once.  100s of models which lack file structure, naming conventions, tests, docs or even refs.
Allow users to specify a clean slate for future-state.  Mainly a CI use case.

### Are you interested in contributing this feature?
Sure!

rsanjabi:
+1 ing 

We're currently getting around legacy code that breaks rules by enabling/disabling rules for the entire repo until we are able to refactor and get it to pass. But having more sophisticated approaches for evaluation would be helpful.

b-per:
Hi @jschintz-windriver and @rsanjabi 

In case you want to see if this would work for your use case you could try out the version in [feature-allow-excluding-models-packages](https://github.com/dbt-labs/dbt-project-evaluator/tree/feature-allow-excluding-models-packages) (not released to the Hub yet).

Any feedback would be great!

jschintz-windriver:
Thanks @b-per .  This worked well for me.
I appreciate your team building this feature.  And the project-evaluator package in general has gretaly helped herd my stubborn cats.

PR #336, opened 2023-05-02
Feat/ignore packages

### LINKED ISSUE #307: allow users to limit output of models by package
### Describe the feature

If i am a user of a package that contains models, i may want to exclude those models from DPE checks. Right now, doing so would require use of the exceptions seed, with entries for each combination of model-to-exlude + fct_model_name to make sure not to run any check on any external model (or use of wildcards when that PR is closed)

it would be easier to just declare what set of packages to consider, and have the logic in the models take care of the rest!

Seed only

### Additional context

[related thread](https://getdbt.slack.com/archives/CU4MRJ7QB/p1676667752771069?thread_ts=1676661048.963109&cid=CU4MRJ7QB)

### Who will this benefit?

- package users
- multi-project deployments in the future

of course i am, me

epapineau:
This is a:
- [x] new functionality

## Link to Issue 

Closes #307

## Description & motivation

I want to enable more dbt_project_evaluator rules in my project, but we have several package dependencies that are also being flagged by the tests. This PR introduces the project variable `include_package_models` which, when `false`, adds a filter to all fct models for the current project's `project_name` jinja variable.

For fact models that get data only from `int_all_dag_relationships`, there is both a `parent_package_name` and a `child_package_name`. I did my best to infer which to alias as `package_name` and filter by:

1. If the final CTE selected exclusively parent or child, keeping this consistent, or
2. Aliasing the column that the dbt developer could manipulate to get the rule into a passing state.

I am confident with the changes, but would appreciate 👀 from someone more familiar with each of the models to confirm my work.

I was originally interested in adding this feature to get the unused sources rule running in my project & I used that model specifically for validation. I ran the exact same set of code against the current version of `dbt_project_evaluator` in dbt Cloud and my local version of the package & project:
Screenshot 2023-05-01 at 6 38 36 PM
Screenshot 2023-05-01 at 7 16 43 PM

Gonna rely on the PR for these coz I couldn't find instructions for running locally in the README or on the project site.
## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
    - [ ] Databricks
    - [ ] DuckDB


b-per:
Thanks @epapineau . It looks like what you did would work well, but I am just wondering if we wouldn't want to make the "exclusion" behaviour more generic to tackle other use cases as well (and in that case merging this would maybe not play well with the other use cases)

e.g., [this issue](https://github.com/dbt-labs/dbt-project-evaluator/issues/234), to allow selecting only some models based on path/name, is tagged to me and I have been wanting to tackle it without finding the time.

The idea we discussed would be to potentially tag at the staging layer level if a model should be excluded or not (e.g. based on the package name, its model name, its path...) in a column called something like `exclude` and then, in the filtering logic, check if the model should be excluded or not.

If the current beahviour is a blocker to use the package internally, I can prioritize this work on developing the "generic exclusion" method.

epapineau:
>I am just wondering if we wouldn't want to make the "exclusion" behaviour more generic to tackle other use cases as well 
Would the idea be ignoring the full contents of the `dbt_packages` directory? Otherwise, I could see a model-by-model approach becoming very cumbersome to configure in a project with several package dependencies like our's.

>If the current beahviour is a blocker to use the package internally, I can prioritize this work on developing the "generic exclusion" method.

This is the case for us. We are implementing and cleaning up our project on a rule-by-rule basis, but we will not be refactoring to an extent to remove dependencies on projects. As such, we are only able to run two rules in our project today without making a massive exclude seed. A massive exclude seed has been ruled out as an option for us because it creates too much work when introducing a new package. Hope that makes sense.

b-per:
Closing in favor of #340 that is a bit more generic.

This is in `main` already but we will do a release soon.

b-per:
Thanks @epapineau for getting us to finally focus on this feature 🙌

Issue #332, opened 2023-04-23
New Rules! for model governance (dbt Core v1.5)


jtcohen6:
### Describe the feature

https://dbt-labs.github.io/dbt-project-evaluator/0.5/rules/

I'd like to propose some [New Rules](https://dbt-labs.github.io/dbt-project-evaluator/main/rules/), related to the model governance features that are new in v1.5.
- **Public Models Have Contracts:** Every model with `access: public` also has `contract: {enforced: true}`
- **Public Models Are Documented:** Every model with `access: public` has a top-level `description` _and_ `description` for every column ([this rule](https://dbt-labs.github.io/dbt-project-evaluator/main/rules/documentation/#undocumented-models), but more)
- **Column Naming Convention:** For columns in models with explicitly defined contracts, we can assert that columns' names and data types match [our opinionated recommendations](https://github.com/dbt-labs/corp/blob/main/dbt_style_guide.md#:~:text=Date/time%20columns,is_active_customer%20and%20has_admin_access).
- **Exposures Public Parents:** Exposures depend only on models with `access: public` (similar to [this rule](https://dbt-labs.github.io/dbt-project-evaluator/main/rules/performance/#exposure-parents-materializations))
- **At Most Three Versions:** There should be, at most, three versions of a given model: one `latest`, one `old`, one `prerelease`
- **Group Coverage:** Calculate percentage of resources in `groups`. Groupable resources are everything except `sources` + `exposures`. Raise a warning about resources that aren't in groups, with an owner. (This will be a heavier lift at first; we should build [tooling to make it easier](https://github.com/dbt-labs/dbt-meshify/issues/21)!)
- **Model Access Ratio:** Calculate ratios of model `access`: private, protected, public. My absolutely arbitrary guess at an ideal breakdown would be somewhere in the ballpark of 50/40/10%. Every `group` should have at least one `private` model.

Related to `deprecation_date` (https://github.com/dbt-labs/dbt-core/issues/7433), coming in v1.6:
- All `old` model versions (< `latest_version`) must have a `deprecation_date`
- All models with a `deprecation_date` earlier than today should be removed

Very open to thoughts & feedback on all of the above — and more ideas, of course :)

### Who will this benefit?

Developers & maintainers of large & complex projects, who want to start taking advantage of v1.5 features for model governance

### Are you interested in contributing this feature?

I can help with refining/brainstorming, &/or possibly contribute with some of my hacktime :)

[...]

PR #339, opened 2023-05-03
add 1.5 features to stg_nodes, fct_public_models_without_contracts, docs

dave-connors-3:
This is a:
- [x] new functionality

## Link to Issue 

This PR is in partial fulfillment of #332 

## Description & motivation

this PR adds:
- access
- contract.enforced
- latest_version
- group
- # of yml cols
- # of yml cols with description

for models and adds:

- `fct_public_models_without_contracts` - outputs each model with `access: public` and `config.contract.enforced: false`

image

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [x] DuckDB


b-per:
Did you check the docs that get generated? I think that you will need to [update the `nav` config](https://github.com/dbt-labs/dbt-project-evaluator/blob/e3fa5f639f4bcf19204587281988cada3611c5e2/mkdocs.yml#L64-L80) because you added a new page.

docs/rules.md
```diff
@@ -29,3 +29,4 @@ hide:
 |Structure                                     |[Model Directories](../rules/structure/#model-directories)                                                             |`fct_model_directories`|
 |Performance                                   |[Chained View Dependencies](../rules/performance/#chained-view-dependencies)                                           |`fct_chained_views_dependencies`|
 |Performance                                   |[Exposure Parents Materializations](../rules/performance/#exposure-parents-materializations)                           |`fct_exposure_parents_materializations`|
+|Governance                                    |[Public Models Without Contracts](../rules/governance/#public-models-without-contracts)                                |`fct_public_models_without_contracts`|
```
b-per, line 32: Aren't there a few more to add here?

b-per (commented):
A couple of changes required for the docs but except that it's all good for me

docs/rules/governance.md
```diff
@@ -0,0 +1,149 @@
...
+models:
+  - name: report_1
+    description: very important OKR reporting model
+    access: public
+
+```
+
+**Reason to Flag**
+
+Models with public access are free to be consumed by any downstream consumer. This implies a need for better guarantees around the model's data types and columns. Adding a contract to the model will ensure that the model *always* conforms to the datatypes you expect.
```
graciegoheen, line 24:
```suggestion
Models with public access are free to be consumed by any downstream consumer. This implies a need for better guarantees around the model's data types and columns. Adding a contract to the model will ensure that the model *always* conforms to the datatypes, columns, and other constraints you expect. 
```

docs/rules/governance.md
```diff
@@ -0,0 +1,149 @@
...
+
+Edit the yml to include the contract configuration, as well as a column entry for all columns output by the model, including their datatype. While not strictly required for defining a contracts, it's best practice to also document each column as well. 
+
+```yml
+models:
+  - name: report_1
+    description: very important OKR reporting model
+    access: public
+    config:
+      materialized: table
```
graciegoheen, line 36: Are public models required to be materialized as a table? If so, let's add a note about that. If not, let's remove this from the example because it's a difference between the public model without a contract example

docs/rules/governance.md
```diff
@@ -0,0 +1,149 @@
...
+        enforced: true
+    columns:
+      - name: id 
+        description: the primary key of my OKR model
+        data_type: integer
+```
+
+## Undocumented public models
+
+`fct_undocumented_public_models` ([source](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/models/marts/governance/fct_undocumented_public_models.sql)) shows each model with `access` configured as public that is not fully documented. This check is similar to `fct_undocumented_models` ([source](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/models/marts/documentation/fct_undocumented_models.sql)), but is a stricter check that will highlight any public model that does not have a model-level description as well descriptions on each of its columns.
```
graciegoheen, line 47: Curious the reason to have _both_ this model and `fct_undocumented_models`? Seems a bit redundant to have both, though this captures documenting at the _column_ level so maybe it's ok for the bit of overlap?

docs/rules/governance.md
```diff
@@ -0,0 +1,149 @@
...
+        data_type: integer
+```
+
+## Undocumented public models
+
+`fct_undocumented_public_models` ([source](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/models/marts/governance/fct_undocumented_public_models.sql)) shows each model with `access` configured as public that is not fully documented. This check is similar to `fct_undocumented_models` ([source](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/models/marts/documentation/fct_undocumented_models.sql)), but is a stricter check that will highlight any public model that does not have a model-level description as well descriptions on each of its columns. 
+
+**Example**
+
+`report_1` is defined as a public model, but does not descriptions on the model and each column
```
graciegoheen, line 51:
```suggestion
`report_1` is defined as a public model, but does not have descriptions on the model and each column 
```

docs/rules/governance.md
```diff
@@ -0,0 +1,149 @@
...
+
+```
+
+**Reason to Flag**
+
+Models with public access are free to be consumed by any downstream consumer. This implies a need for higher standards for the model's usability for those cosumers. Adding more documentation can help consumers understand how they should leverage the data from your public model.
+
+**How to Remediate**
+
+Edit the yml to include a model level description,  as well as a column entry with a description for all columns output by the model. While not strictly required for public models, these should likely also have contracts added as well. (See above rule)
```
graciegoheen, line 69: Can you link to the "above" rule (and name it, in case at some point it's no longer "above")

docs/rules/governance.md
```diff
@@ -0,0 +1,149 @@
...
+
+```yml
+models:
+  - name: report_1
+    description: very important OKR reporting model
+    access: public
+    config:
+      materialized: table
```
graciegoheen, line 77: same note as earlier about `materialized: table`

docs/rules/governance.md
```diff
@@ -0,0 +1,149 @@
...
+## Undocumented public models
+
+`fct_undocumented_public_models` ([source](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/models/marts/governance/fct_undocumented_public_models.sql)) shows each model with `access` configured as public that is not fully documented. This check is similar to `fct_undocumented_models` ([source](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/models/marts/documentation/fct_undocumented_models.sql)), but is a stricter check that will highlight any public model that does not have a model-level description as well descriptions on each of its columns. 
+
+**Example**
+
+`report_1` is defined as a public model, but does not descriptions on the model and each column 
+
+```yml
+# public model without a contract
```
graciegoheen, line 54: Should this comment be different for this example? `#public model that's not documented`

docs/rules/governance.md
```diff
@@ -0,0 +1,149 @@
...
+Edit the yml to include a model level description,  as well as a column entry with a description for all columns output by the model. While not strictly required for public models, these should likely also have contracts added as well. (See above rule)
+
+```yml
+models:
+  - name: report_1
+    description: very important OKR reporting model
+    access: public
+    config:
+      materialized: table
+      contract:
```
graciegoheen, line 78: This example shows the fix as also adding a contract. I wonder if we could have 2 different examples for these 2 rules so it's clear what needs to be added to remediate this specific issue

docs/rules/governance.md
```diff
@@ -0,0 +1,149 @@
...
+        enforced: true
+    columns:
+      - name: id 
+        description: the primary key of my OKR model
+        data_type: integer
+```
+
+## Exposures dependent on private models
+
+`fct_exposures_dependent_on_private_models` ([source](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/models/marts/governance/fct_exposures_dependent_on_private_models.sql)) shows each relationship between a resource and an exposure where the parent resource is not a model with `access` configured as public.
```
graciegoheen, line 88: Do we still need `fct_exposure_parents_materializations` or can these be combined?

docs/rules/governance.md
```diff
@@ -0,0 +1,149 @@
...
+
+**Example**
+
+Here's a sample DAG that shows direct exposure relationships. 
+
+![An example exposure with a two parents (fct_model_6 and dim_model_7)](https://user-images.githubusercontent.com/73915542/178068955-742e2c87-4385-48f9-b9fb-94a1cbc8079a.png){ width=500 }
+
+If this were the yml for these two parent models, `dim_model_7` would be flagged by this model, as it is not a public model.
```
graciegoheen, line 96:
```suggestion
If this were the yml for these two parent models, `dim_model_7` would be flagged by this check, as it is not a public model. 
```

docs/rules/governance.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+
+**Reason to Flag**
+
+Exposures show how and where your data is being consumed in downstream tools. These tools should read from public, trusted, contracted data sources. All models that are exposed to other tools should have that codified in their `access` configuration. 
+
+**How to Remediate**
+
+Edit the yml to include fully expose the models that your exposures depend on. This should include `access`, a full model contract, and descriptions at all levels of the model.
```
graciegoheen, line 123: Again, what is the _actual_ rule here? If it's just checking for access... should we also show adding contracts, descriptions, etc.

macros/recursive_dag.sql
```diff
@@ -31,6 +33,8 @@ all_relationships (
     child_resource_type,
     child_model_type,
     child_materialized,
+    child_access,
+    child_is_public,
```
graciegoheen, line 37: Why the need for both `access` and `is_public` - isn't `is_public` just `if access == 'public'`?

docs/rules/governance.md
dave-connors-3: I think it's to have a stricter requirement that is called out separately (better signal to noise ratio) and also have details at the column level as you note. It's more important for these models to be documented than it is for a random staging model!

docs/rules/governance.md
dave-connors-3: done!

docs/rules/governance.md
dave-connors-3: yeah i note in the description that it's not required, but encouraged -- I can remove if we think it's too confusing

docs/rules/governance.md
dave-connors-3: it's a good question! this one is a bit stricter -- do we think folks not using governance features will get value out of this rule in the absence of the other?

macros/recursive_dag.sql
dave-connors-3: I think just to allow for future rules that use `access`! can remove for now if we don't think it's valuable

docs/rules/governance.md
dave-connors-3: unless this is a create or replace issue 🤔

PR #345, opened 2023-05-15
adjust root models to only filter excluded children nodes

### LINKED ISSUE #344: Models with package dependencies raise errors in root models rule
### Describe the bug

I am implementing the package exclusion feature added in #340.  The `fct_root_models` model test is failing for two models in our project with have dependencies on source or models from a package. These models do not actually violate the root model rule, so I don't want to add them to my exceptions seed.

### Steps to reproduce

Use a package `source` or `ref` in your main project and execute `dbt build -s fct_root_models`
### Expected results

If the dependency is from a package, it will not trigger a test failure.

### Actual results

The models are 
### System information

- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [x] snowflake

```
<output goes here>
```
We're on 1.4.6

### Additional context

From @graciegoheen:
> seems like we need to update [this](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/models/marts/dag/fct_root_models.sql) model to only filter out where the child is excluded - but keep all parents

graciegoheen:
This is a:
- [ ] new functionality

## Link to Issue 

Closes #344 

## Description & motivation

For `fct_root_models`, we should only filter out excluded children nodes (_not_ parents) - filtering parents could result in incorrectly flagging nodes that depend on excluded nodes.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB

### DIFF
--- a/models/marts/dag/fct_root_models.sql
+++ b/models/marts/dag/fct_root_models.sql
@@ -5,7 +5,8 @@ with model_relationships as (
         *
     from {{ ref('int_all_dag_relationships') }}
     where child_resource_type = 'model'
-    and not parent_is_excluded
+    -- only filter out excluded children nodes
+        -- filtering parents could result in incorrectly flagging nodes that depend on excluded nodes
     and not child_is_excluded
 ),
 


epapineau (approved):
Confirming the resolves the issue I experienced leading to #344

dave-connors-3 (approved):

b-per:
I don't know if we are fixing the real issue here or if this is just a workaround for this particular use case. 

Without checking further yet, my first question is why deactivating a source package would raise some entries in `fct_root_models`?

PR #350, opened 2023-05-26
🐛 Fix exceptions, by reading `model.name`

### LINKED ISSUE #341: Question: get_custom_alias() with filter_exceptions()
I have been testing this package and am loving it. 

Our project overrides the `get_custom_alias macro()` to support multiple developers in one schema. 

For example, if I build` foo.sql` in our `dev` schema, it would get built as `dev.dbt_ml_foo`.

I have exceptions listed out in the `dbt_project_evaluator_exceptions.csv`.  The issue I've run into, is that when using the exceptions, the macro `filter_exceptions()` is using the model name per the target being run, as it's built in the warehouse (in other words, it's searching the exception file for `dbt_ml_fct_model_naming_conventions` instead of `fct_model_naming_conventions`). 

Instead, I want it to strip off my target prefix when using that macro. 

For the time being, I modified the macro, like this 👇 , essentially just replacing `dbt_ml` with an empty string so my exception can be found in the seed file. 
image

But, I wanted to see if there's a different way to go about this without modifying the macro. Or I'm even happy to contribute a customization for this macro if that's the best path forward.

b-per:
`this.name` gives the table name and not the model name

This is a:
- [ ] new functionality

## Link to Issue 

Fixes #341

## Description & motivation

TIL that `this.name` returns the table/view name (a.k.a with `generate_alias_name()` applied), when `model.name` just returns the model name.

This is a breaking change (in case some people where using an updated `generate_alias_name()` and defining exceptions on those), but the updated behavior is the correct one.

### DIFF
--- a/macros/filter_exceptions.sql
+++ b/macros/filter_exceptions.sql
@@ -1,15 +1,15 @@
-{% macro filter_exceptions(model_ref) -%}
-    {{ return(adapter.dispatch('filter_exceptions', 'dbt_project_evaluator')(model_ref)) }}
+{% macro filter_exceptions(model_name) -%}
+    {{ return(adapter.dispatch('filter_exceptions', 'dbt_project_evaluator')(model_name)) }}
 {%- endmacro %}
 
-{% macro default__filter_exceptions(model_ref) %}
+{% macro default__filter_exceptions(model_name) %}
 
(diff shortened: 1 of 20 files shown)


dave-connors-3 (approved):
🦞

PR #362, opened 2023-07-19
only consider sql models in find_all_hard_coded_references

### LINKED ISSUE #357: Hard Coded References interaction with dbt-python (Snowpark)
## Context
When using [dbt-python](https://docs.getdbt.com/docs/build/python-models) there is usually the necessity of importing packages.

## Problem
In the `fct_hard_coded_references` table I get a list of all the packages imported as an "error". 
For example:
- from {package} import {function} generates an entry on the hard_coded_references list, indicating {package}

## Desired Output
It would be nice to skip python imports when executing this check

dave-connors-3:
This is a:
- [ ] new functionality

## Link to Issue 
Closes #357

## Description & motivation

python models that use `from {{ package }} import {{ some stuff }}` syntax are getting flagged for using hard coded references! We should only run this check on sql models

The addition of a python model to the integration test suite is definitely possible, but would require having adapter specific settings to only run where python models exist, and a bunch of other edits to our seed files to keep it all in check. This feels low risk enough to me that I decided against all that, but can definitely reconsider if we decide this should be explicitly tested

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [x] DuckDB

### DIFF
--- a/macros/find_all_hard_coded_references.sql
+++ b/macros/find_all_hard_coded_references.sql
@@ -6,7 +6,8 @@
 
     {%- set all_hard_coded_references_list = [] -%}
 
-    {% if node.resource_type == 'model' %}
+    {% if node.resource_type == 'model' and node.language == "sql" %}
+        {{ log(node.language, info=True) }}
 
         {% if execute %}
         {%- set model_raw_sql = node.raw_sql or node.raw_code -%}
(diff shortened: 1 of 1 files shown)


graciegoheen (approved):
I like this as a short term solution. We can consider expanding this to python models in the future, but regex can only cover so much!

b-per (approved):

PR #366, opened 2023-08-04
Update broken links in docs

ttusing:
This is a:
- [X] bug fix PR with no breaking changes
- [ ] new functionality

## Description & motivation
These links are broken. For some reason, the relative paths do not work as I would anticipate. Navigating from `readme.md`, the link is `https://dbt-labs.github.io/dbt-project-evaluator/`, which forwards me to `https://dbt-labs.github.io/dbt-project-evaluator/0.6/`. Clicking on `modeling` sends me to `https://dbt-labs.github.io/rules/modeling` which does not exist. A valid page includes the version, such as `https://dbt-labs.github.io/main/rules/modeling`. I believe the attached changes should send users to an expected page. There may be a smarter way to manage these links to match the version of the home page you've landed on.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -3,11 +3,11 @@
 This package highlights areas of a dbt project that are misaligned with dbt Labs' best practices.
 Specifically, this package tests for:
 
-1. __[Modeling](./rules/modeling)__ - your dbt DAG for modeling best practices
-1. __[Testing](/rules/testing)__ - your models for testing best practices
-1. __[Documentation](/rules/documentation)__ - your models for documentation best practices
-1. __[Structure](/rules/structure)__ - your dbt project for file structure and naming best practices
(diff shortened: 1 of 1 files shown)


b-per:
Thanks for raising the issue and making a PR for it.

The suggested solution would always point to `main` though, even if someone selected another version. I think that we just need to make the paths relative rather than absolute (some info [here](https://github.com/jimporter/mike/issues/20)).

I will have a look at it.

b-per:
Thanks again for raising the issue. I just merged #367 to fix the links but still making it work for different versions.

PR #376, opened 2023-09-13
Add support for Trino

### LINKED ISSUE #375: Add support for Trino
### Describe the feature
Add support for Trino, so this package could be used with [dbt-trino](https://github.com/starburstdata/dbt-trino) 

Initially, we considered adding dbt-project-evaluator to [dbt-trino-utils package](https://github.com/starburstdata/dbt-trino-utils), which defines implementations of dispatched macros from other packages that can be used on Trino or Starburst.
However, we realised that it would be quite tedious to override all macros and models (and descendant models) on our side. It would also interfere with already existing packages in dbt-trino-utils. This approach would also require users to install dbt-trino-utils and make additional configurations in their dbt_project.yml file.

### Additional context
This is relevant for Trino.

### Who will this benefit?
Users who want to use dbt-project-evaluator with [dbt-trino](https://github.com/starburstdata/dbt-trino) 

Yes!

damian3031:
This is a:
- [x] new functionality

## Link to Issue 
Closes #375 

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB
    - [x] Trino


damian3031:
I've added integration tests to CircleCI. Please add below environment variables relevant for your Starburst Galaxy instance (AFAIK you have one) so that the tests can run on it.
`TRINO_TEST_HOST`, `TRINO_TEST_USER`, `TRINO_TEST_PASS`, `TRINO_TEST_PORT`, `TRINO_TEST_CATALOG_NAME`

b-per:
Thanks @damian3031 ! 
I am testing it locally right now and we will add the Starburst credentials for CI/CD!

Based on my initial testing (using a Snowflake catalog) there are some issues (which might required updates to `dbt-trino`):
- if a table has a description, it is added to the DDL but I then get the error `TrinoUserError(type=USER_ERROR, name=NOT_SUPPORTED, message="This connector does not support creating tables with table comment", query_id=abcd)`
  - is it possible to disable the comment part of the DDL? it doesn't seem to be based on the code I saw in `dbt-trino`
- I am also getting some errors `This connector does not support creating views` which seems consistent with [this issue](https://github.com/trinodb/trino/issues/11588).

Based on the observations above:
- what catalog type would you recommend to avoid those errors? (or which one did you test with for now)
- should we test the package for different catalog types?
- we might need to add some info in the README to mention that some catalog types will work with Trino but not all

damian3031:
@b-per creating views is supported in object storage connectors such as Hive, Delta Lake, and Apache Iceberg.

I tested it on the Iceberg connector, and the tests passed.

I think it would be fair to test it specifically against Iceberg connector and mention it in the README.

README.md
```diff
@@ -20,6 +20,7 @@ Currently, the following adapters are supported:
 - Redshift
 - Snowflake
 - DuckDB
+- Trino
```
damian3031, line 23:
```suggestion
- Trino (tested with Iceberg connector)
```
@b-per Would that be ok?

.circleci/config.yml
```diff
@@ -104,3 +115,4 @@ workflows:
           context: profile-databricks
           requires:
             - integration-postgres
+      - integration-trino
```
b-per, line 118: Could you please just add a 
```
          requires:
            - integration-postgres
```
like the other Cloud Warehouses? The idea is that we try first with the "offline" ones so we don't incur any Cloud cost if those fail.

b-per (changes requested):
Thanks for the work here!

All good with me with the small exception of updating the CI config file to run Trino at the same time as Snowflake/BQ etc...

I have added the env vars to test on S3 and it works

.circleci/config.yml
damian3031: Done!
I've also added `context: profile-trino`

b-per (approved):

b-per:
We'll do a new release of the package on the dbt Hub in just a beat so that it is accessible to users.

PR #377, opened 2023-09-18
Epapineau/undocumented sources

### LINKED ISSUE #374: Report on undocumented sources
### Describe the feature
A parallel to `fct_undocumented_models` where `resource_type = 'source'`
We currently use dbt-check for this and only this. I'd love to retire it and go all in on project evaluator 🎊 

### Additional context
Is this feature database-specific? Which database(s) is/are relevant? Please include any other relevant context here.

### Who will this benefit?
What kind of use case will this feature be useful for? Please be specific and provide examples, this will help us prioritize properly.
dbt developers and data engineers. The source is the beginning of the dbt project and one of the most important important places to communicate caveats around loading strategy.

maybe a lil

epapineau:
This is a:
- [x] new functionality

## Link to Issue 

Closes #374

## Description & motivation

This PR adds documentation testing for both source- and source-node-level as separate entites. 

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
    - [ ] Databricks
    - [ ] DuckDB


graciegoheen:
What is the difference between a "source" and a "source node"? Why have them as separate checks?

epapineau:
@graciegoheen

The difference is the source level or table level, which can both have descriptions, unlike models. For example:
```
sources:
  - name: my_source
    description:  This is a schema-grain description
    tables:
      - name: my_table
        description: This is a table-grain description
 ```

With the "table-grain" being called a source node. I assumed it would be preferable to keep them separate to emphasize the different grains, but I am happy to combine them if that is what y'all want to see.

**Comment by b-per** [comment]
I know that sources could be views and not just tables, (hence why node might be more appropriate), but to keep the same taxonomy as in the source YML files, could we call the 2 different levels `source` and `source_table`

File: docs/rules/documentation.md [hunk 1]
```diff
@@ -46,3 +46,29 @@ function in the model's `.yml` entry.
 !!! note "Tip"
 
     We recommend that every model in your dbt project has at minimum a model-level description. This ensures that each model's purpose is clear to other developers and stakeholders when viewing the dbt docs site. Missing documentation should be addressed first for marts models, then for the rest of your project, to ensure that stakeholders in the organization can understand the data which is surfaced to them.
+
+## Undocumented Source Nodes
```
dave-connors-3, line 50: i think it would be good to include the little sample yml snippet you used in the conversation on this PR to clarify which description block would resolve each of these rules!

epapineau: Done!

epapineau:
@b-per, @dave-connors-3 Thanks for the feedback! I believe I've integrated all of your recommendations

integration_tests/seeds/docs/test_fct_undocumented_sources.csv
```diff
@@ -0,0 +1,2 @@
+resource_name
+source_2.table_3
```
dave-connors-3, line 2: @epapineau in terms of what we want to display, would it make sense to either strip this string to `source_2` rather than including the table name too? we may have that as a column in the resources table already! Not blocking, but may clarify what the user needs to do to clear this check

models/marts/documentation/fct_undocumented_sources.sql
```diff
@@ -0,0 +1,22 @@
...
+all_resources as (
+    select * from {{ ref('int_all_graph_resources') }}
+    where not is_excluded
+
+),
+
+final as (
+
+    select
+        resource_name,
```
dave-connors-3, line 12: per the above, we can use `all_resource.source_name`, then make this a `distinct`

dave-connors-3 (changes requested):
small change! for clarity in `fct_undocumented_sources`, I think we should return only the parent source name, and not the table names! should be a quick fix!

integration_tests/seeds/docs/test_fct_undocumented_sources.csv
epapineau: 💯

models/marts/documentation/fct_undocumented_sources.sql
epapineau: Brilliant - thank you

epapineau:
@dave-connors-3 I've integrated your suggestions

dave-connors-3 (approved):
🦞

docs/rules.md
```diff
@@ -23,6 +23,8 @@ hide:
 |Testing                                       |[Test Coverage](../rules/testing/#test-coverage)                                                                       |`fct_test_coverage`|
 |Documentation                                 |[Undocumented Models](../rules/documentation/#undocumented-models)                                                     |`fct_undocumented_models`|
 |Documentation                                 |[Documentation Coverage](../rules/documentation/#documentation-coverage)                                               |`fct_documentation_coverage`|
+|Documentation                                 |[Undocumented Source Tables](../rules/documentation/#undocumented-source-tables)                                                     |`fct_undocumented_source_tables`|
+|Documentation                                 |[Undocumented Sources](../rules/documentation/#undocumented-sources)                                               |`fct_documentation_sources`|
```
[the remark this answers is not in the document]
graciegoheen: https://github.com/dbt-labs/dbt-project-evaluator/pull/382

Issue #199, opened 2022-09-09
Documentation evaluation: column level and invalid


elyobo:
### Describe the feature

The existing documentation checks only pick up undocumented models, but there are some extensions that would be useful.

* invalid documented models, e.g. where the documented model does not exist (perhaps because it has been removed or renamed but the documentation has not)
* undocumented columns, as per the existing model functionality but per column instead
* invalid documented columns, as per the "invalid documented models" suggestion above but per column instead

Per column level warnings are likely to be very verbose, so fine grained control over what is reported would be necessary, and defaulting the tests to be disabled would probably be appropriate - opt in, not opt out.

### Describe alternatives you've considered

I have seen (although I can't find them now) cleverly constructed tests that handle the undocumented case in a very heavy way - IIRC the tests needed to be defined per model and/or per column and so would infer the non-existence of the model or column by running invalid queries.

I have not seen anything for identifying undocumented columns elsewhere.

### Additional context

Presumably not DB specific.

### Who will this benefit?

Same target audience as the other documentation checks, those wanting a good level of documentation coverage across their data structures.

### Are you interested in contributing this feature?

I would, but have no idea where to start (more a JS dev than a python dev).

dave-connors-3:
Hey @elyobo! Thanks for opening the issue! 

for situations where there is documentation for models or columns that don't exist, I believe that dbt will throw a `[WARNING]` message at the beginning of each dbt command. Is this the behavior you'd also want to build a check into this package for? 

Column level documentation coverage is a really cool idea -- I don't think we currently have any checks built out at the column level at the moment. What would your ideal output of a model that checks this look like? A row per each column that is not documented? or one per model with a list of the undocumented columns? 

cc: @graciegoheen for visibility!

elyobo:
Can confirm that models that don't exist are warned about (so that feature is probably redundant, unless wanted for consistency) but columns don't appear to be.

Example model warning below, no warning given for the `this_column_does_not_exist` column I added a description to on a model that _does_ exist but does not have that column.

```
23:47:39  [WARNING]: Did not find matching node for patch with name 'this_momdel_does_not_exist' in the 'models' section of file 'models/report/_report__models.yml'
```

I was thinking row per column, similar to undocumented models.

PR #386, opened 2023-10-07
feat: Inspect columns

### LINKED ISSUE #385: Column-level granularity
### Describe the feature
Hello guys, thank your for working on this amazing package. We are trying to use it as our only tool when it comes to DBT project validation. In order to achieve this, we need to run certain custom validation rules on the top of columns, not just nodes (models). To give you an example, we would like to forbid certain column names to be used across our project.

We hoped, that `int_all_graph_resources` is going to give as this answer by seeing `column_name` column in it, but it looks like it's set only for generic tests.

Then we thought, maybe we can find anything in the underlaying staging models like `stg_nodes`, but still no luck.

Lastly, we check the underlaying implementation:
https://github.com/dbt-labs/dbt-project-evaluator/blob/601263c23127efb53678110d45c5c5148b44ec1d/macros/unpack/get_node_values.sql#L31-L32
And it's clear you gather only information related to columns count, not what they are.

Additionally, there was a request last year related to validating missing column descriptions, which was automatically closed:
https://github.com/dbt-labs/dbt-project-evaluator/issues/199

### Who will this benefit?
Anyone who is willing to start using column-level rules.

I gave it a go: https://github.com/dbt-labs/dbt-project-evaluator/pull/386

gastlich:
This is a:
- [X] new functionality

## Link to Issue 
Closes #385 #199 

## Description & motivation
As mentioned in #385, it would be great if this tool could create a model using all the information gathered for columns. This would enable us to begin implementing basic validation rules for columns, such as:

1. Checking for missing descriptions.
2. Ensuring correct naming conventions for columns, which would prevent certain names from being used.

Furthermore, in the future, we can expand its use for more advanced validation purposes.

This is my first PR in this project and it would be good if you shared with me some hints/suggestions how to deliver the best outcome.

## TODO
- [x] Integration Tests
- [ ] Test on postgres
- [x] Check if `sources.tables` changes data structure from `graph`
- [x] Update documentation

## Checklist
- [X] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [X] BigQuery Screenshot 2023-10-07 at 20 26 03
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB
    - [ ] Trino/Starburst


dave-connors-3:
hey @gastlich!

this is looking great, really appreciate your work here! I was able to run this locally and is seems to be working great! I think this is a great compromise to provide folks with a base to build column level checks into their projects without creaing unnecessary noise for those who don't want or need column level checks!

Couple small things:
 1. could you add tests and descriptions to the new models (at least to `stg_columns`?) Would be great to ensure the grain is consistent and have clear docs to explain the purpose of the model
 2. could you add a blurb about this model set into the actual documentation website? I think maybe a new markdown file in the [docs/customization](https://github.com/dbt-labs/dbt-project-evaluator/tree/main/docs/customization) folder could be appropriate (and a corresponding addition to the [config](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/mkdocs.yml#L78)) 

Let me know if you need a hand with either of these!

gastlich:
hey @dave-connors-3 

Thanks for you feedback and suggestions! I've pushed some changes including:


1. I've added definitions for `stg_columns.sql` to `graph.yml` and `int_all_columns.sql` to `core.yml`. I hope this is what you meant by "adding descriptions to the new models"
2. When it comes to "consistent" grain, would you prefer to use surrogate key, which is a new column with the value defined as `<node.unique_id> + <column.name>` or composite key with test implemented with `dbt_utils.unique_combination_of_columns`?
3. I would like to ask you what's your preference regarding the naming of columns in the new models? Should I prefix all of the node's columns with `node_` prefix to make it explicit, that it doesn't relate to column? For example `node_unique_id` instead of `unique_id`, and `node_name` instead of `name`?
4. I've added a new documentation page `querying-columns.md` with a basic explanation of the model.
5. I updated `mkdocs.yml` to include the new page.

I still need to figure out what's needed for the integration tests, but this is a task for another evening :) Could you review what has been recently published, please?

models/staging/graph/stg_columns.sql
```diff
@@ -0,0 +1,19 @@
+{{
+    config(
+        materialized='table',
+    )
+}}
+
+{% if execute %}
```
dave-connors-3, line 7: i think techincally speaking, this one doesn't need to be a table! that check is important for the tables where we're abusing the `post_hook` feature to run some insert statements, which can't be done on views!

dave-connors-3 (changes requested):
this is looking really good! Given that this package runs on some really large customer projects, and therefore there's a risk that this model might be rather non-performant, I think I'd prefer to have the columns models disabled by default, and allow users to turn them on if they want to run them. Would you mind adding those configs to the `dbt_project.yml`? 

Other than that, just a couple small suggestions!

docs/customization/querying-columns.md
```diff
@@ -0,0 +1,16 @@
+# Querying columns with SQL
+
+The model `int_all_columns` ([source](https://github.com/dbt-labs/dbt-project-evaluator/tree/main/models/marts/core/int_all_columns.sql)), created with the package, lists all the columns from all the dbt nodes (models, sources, seeds, snapshots, analyses)
+
+You can use this model to help with questions such as:
+
+- Are there columns with the same name in different nodes?
+- Are there columns that lack descriptions?
```
b-per, line 8: This will only look at columns that have been manually added to the YAML files and not all the columns of every model. Could we re-word this sentence to mention it?

macros/unpack/get_column_values.sql
```diff
@@ -0,0 +1,46 @@
...
+                {%- set values_line  = 
+                    [
+                        wrap_string_with_quotes(node.unique_id),
+                        wrap_string_with_quotes(node.name),
+                        wrap_string_with_quotes(node.resource_type),
+                        wrap_string_with_quotes(node.original_file_path | replace("\\","\\\\")),
+                        wrap_string_with_quotes(node.database),
+                        wrap_string_with_quotes(node.schema),
+                        wrap_string_with_quotes(node.package_name),
+                        wrap_string_with_quotes(node.alias),
```
b-per, line 30: Could we remove those and instead join with the node model? We would just need the `unique_id` to be able to perform this join.

This table potentially having many rows in it, whatever we can do to make is as "small" as possible will help.

b-per (changes requested):
Thanks for the work on this one!

I added a couple of comments about small changes.

b-per:
I just ran some performance tests on our Internal Analytics project wit a significant amount of models

The memory footprint was checked with `usr/bin/time -l dbt ...`, looking at `maximum resident set size`

### Running DPE with the new models
- Memory: 427 MB
- Time: 55s

### Running DPE without the new models
- Memory: 411 MB
- Time: 43s

### Running DPE just for the new models
- Memory: 435 MB
- Time: 24s


## Conclusions
- This new table has no impact on the Memory used by dbt
- Even on a large project, the performance impact is just 10s

@dave-connors-3 , what do you think about disabling the model by default. Does it still make sense? I'd be OK with having it enabled by default, especially if it unlocks further use cases (e.g. checking constraints + tests)

gastlich:
@dave-connors-3 @b-per 

1. I've updated docs
2. Rewritten `int_all_columns` as import CTE, joining with `stg_nodes`
3. Made `stg_columns` a view

Let me know what the next steps are. Thanks a lot for looking into this! 🙇

models/marts/core/int_all_columns.sql
```diff
@@ -0,0 +1,59 @@
...
+        stg_nodes.alias as node_alias,
+        stg_columns.name as name,
+        stg_columns.description as description,
+        stg_columns.data_type as data_type,
+        stg_columns.quote as quote
+
+    from stg_columns
+
+    right join stg_nodes
+        on stg_nodes.unique_id = stg_columns.node_unique_id
```
b-per, line 55: 2 questions/points 😄 

- sorry to be picky, but can we use a `left join`? In our code we try to only use `left` or `inner`/`outer` joins, and never a `right`. This is just to make the mental models of joins easier to understand for someone seeing the code
- why would we want a `left`/`'right`, instead of a `inner` join? Do you want to use this table to see which table doesn't have any column configured in the YAML?

b-per (changes requested):
Thanks a lot for the changes. One last comment from me.

models/staging/graph/stg_columns.sql
```diff
@@ -0,0 +1,11 @@
...
+
+final as (
+
+    {{ dbt_utils.union_relations([
+        ref('base_node_columns'),
+        ref('base_source_columns')
+    ])}}
+)
+
+select distinct * from final
```
kokorin, line 11: `dbt_utils.union_relations` adds synthetic column `_dbt_source_relation`. This makes `select distinct *` ineffective (i.e. it will produce the same result as `select *`).

models/marts/core/int_all_columns.sql
gastlich: Yes, this was my idea to not drop any nodes. But you are right, this is already intermediate layer, thus we can be dropping some data in order to expose higher quality data to the user.

I've changed the join to be inner.

gastlich:
@dave-connors-3 @b-per any other thoughts. I think I still need to add something to the integration tests? 🤔

dave-connors-3:
hey @gastlich! any update on adding tests here? would love to get this in!

gastlich:
Hey @dave-connors-3, Happy New Year! Apologies for the delay; I was quite busy during the Christmas break. I've managed to find some time to work on integration tests.

Just to ensure we're aligned: as I haven't created any `fct_` tests specifically based on the `int_all_columns` model, it's not straightforward to test fragments of that model. Instead, I've provided a single test covering the entire `int_all_columns` model. In the future, when you opt to introduce more specialised column-based tests to this library, you can substitute my generic tests with the newly developed ones.

I hope this meets your requirements. Looking forward to your review! 🙇

integration_tests/models/dbt_project_evaluator_schema_tests/core.yml
```diff
@@ -34,4 +34,12 @@ models:
...
         tests:
           - unique
-          - not_null
\ No newline at end of file
+          - not_null
+
+  - name: int_all_columns
+    description: "This table shows one record for each column for all nodes and sources in the graph."
+    tests:
+      - dbt_utils.unique_combination_of_columns:
```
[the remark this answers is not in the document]
dave-connors-3: (this and the similar test on `stg_columns`)

models/marts/core/int_all_columns.sql
```diff
@@ -0,0 +1,59 @@
...
+        stg_nodes.package_name as node_package_name,
+        stg_nodes.alias as node_alias,
+        stg_columns.name as name,
+        stg_columns.description as description,
+        stg_columns.data_type as data_type,
+        stg_columns.quote as quote
+
+    from stg_columns
+
+    inner join stg_nodes
```
dave-connors-3, line 54: `stg_nodes` does not contain sources, so this inner join is likely removing columns from sources. Is there a need to join in all this node-level info to this view? wondering if it makes more sense to just leave this fairly untransformed, and join in node-level info when needed in future fct models?

macros/unpack/get_column_values.sql
```diff
@@ -0,0 +1,39 @@
...
+            {% set nodes_list = graph.nodes.values() %}   
+        {%- elif node_type == 'sources' -%}
+            {% set nodes_list = graph.sources.values() %}
+        {%- else -%}
+            {{ exceptions.raise_compiler_error("node_type needs to be either nodes or sources, got " ~ node_type) }}
+        {% endif -%}
+
+        {%- set values = [] -%}
+
+        {%- for node in nodes_list -%}
```
dave-connors-3, line 18: per the question about joining, if we wanted to include some node level info, we also have the option of grabbing these values here from the `node` itself.

dave-connors-3 (changes requested):
thanks for your patience @gastlich! a couple things: 

- i think we can get rid of the `core_seeds` stuff -- apologies if i misled you, but we generally just use those seeds to assert the output of the fact models. doing a uniqueness test is probably enough. 
- i think we should rething the join in `int_all_columns` -- we are excluding source oclumn in the current implementation, and I think we can get away with relatively little additional context in that column table, and people can join back on unique_id as needed. I can talk with the team and see if they agree on that last point, but i think the join will still need to be rethought!

integration_tests/models/dbt_project_evaluator_schema_tests/core.yml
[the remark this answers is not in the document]
gastlich: I've removed the equality tests and confirmed that we have `unique_combination_of_columns` for the mentioned models.

models/marts/core/int_all_columns.sql
gastlich: I can leave it untransformed, with the only node-related column coming from `stg_columns`: `node_unique_id`.

I aimed to provide as much information as possible because:
1. I presumed that having One Big Table is preferable in this case to limit potential joins.
2. Besides the missing descriptions and used names, repeated names, I can't think of any other features that depend solely on pure columns' information. Perhaps more sophisticated use-cases will be in relation to a node.
3. In other `int_` models like `int_all_dag_relationships`, we do include parent's/children's data, but we could just as well say: `parent_node_id` and `child_node_id`. Then you could join twice with `int_all_graph_resources` and achieve the same outcome, but for some reason, all the data ended up in this One Big Table `int_all_dag_relationships`.

As mentioned initially, it's up to you to decide the preferred way of modeling this data :)

macros/unpack/get_column_values.sql
gastlich: @dave-connors-3 

Thanks for flagging. It was implemented like this in the past:
https://github.com/dbt-labs/dbt-project-evaluator/pull/386#discussion_r1373413986

But, as suggested by @b-per, we transitioned to the explicit join. I believe it makes more sense because we centralize the extraction of nodes' information in a single place. This ensures we maintain a single source of truth for the node's data and adhere to the DRY principle.

gastlich:
> thanks for your patience @gastlich! a couple things:
> 
>     * i think we can get rid of the `core_seeds` stuff -- apologies if i misled you, but we generally just use those seeds to assert the output of the fact models. doing a uniqueness test is probably enough.
> 
>     * i think we should rething the join in `int_all_columns` -- we are excluding source oclumn in the current implementation, and I think we can get away with relatively little additional context in that column table, and people can join back on unique_id as needed. I can talk with the team and see if they agree on that last point, but i think the join will still need to be rethought!

@dave-connors-3 I've removed the test seed file and `core_seeds.yml`. When it comes to the join, I think you are right, that you need to agree with the rest of the team on what the best approach here should be :) I'm not as familiar with the project, in order to decide on the final look! Let me know, once you have an update.

models/marts/core/int_all_columns.sql
dave-connors-3: hey @gastlich ! we spoke this morning, and decided that we should remove this model, and just leave column information relatively untransformed up to `stg_columns` -- then users  can build and join from there as necessary to create their own rules!

gastlich: @dave-connors-3 I've removed `int_all_columns` model. Changed the documentation and removed any references to the deleted model. Please have a look :)

macros/unpack/get_column_values.sql
```diff
@@ -0,0 +1,39 @@
...
+            {%- for column in node.columns.values() -%}
+
+                {%- set values_line  = 
+                    [
+                        wrap_string_with_quotes(node.unique_id),
+                        wrap_string_with_quotes(dbt.escape_single_quotes(column.name)),
+                        wrap_string_with_quotes(dbt.escape_single_quotes(column.description)),
+                        'null' if not column.data_type else wrap_string_with_quotes(dbt.escape_single_quotes(column.data_type)),
```
graciegoheen, line 26: I believe we can remove the `'null' if not column.data_type else` because the [wrap_string_with_quotes](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/macros/wrap_string_with_quotes.sql) macro automatically handles NULLs

```suggestion
 wrap_string_with_quotes(dbt.escape_single_quotes(column.data_type)),
```

macros/unpack/get_column_values.sql
```diff
@@ -0,0 +1,39 @@
...
+                        'null' if not column.quote else wrap_string_with_quotes(dbt.escape_single_quotes(column.quote))
```
graciegoheen, line 27: ^same as above

```suggestion
                       wrap_string_with_quotes(dbt.escape_single_quotes(column.quote))
```

macros/unpack/get_column_values.sql
graciegoheen, line 26:
```suggestion
                        wrap_string_with_quotes(dbt.escape_single_quotes(column.data_type)),
```

macros/unpack/get_column_values.sql
```diff
@@ -0,0 +1,39 @@
...
+        {%- for node in nodes_list -%}
+            {%- for column in node.columns.values() -%}
+
+                {%- set values_line  = 
+                    [
+                        wrap_string_with_quotes(node.unique_id),
+                        wrap_string_with_quotes(dbt.escape_single_quotes(column.name)),
+                        wrap_string_with_quotes(dbt.escape_single_quotes(column.description)),
+                        wrap_string_with_quotes(dbt.escape_single_quotes(column.data_type)),
+                       wrap_string_with_quotes(dbt.escape_single_quotes(column.quote))
```
graciegoheen, line 27:
```suggestion
                        wrap_string_with_quotes(dbt.escape_single_quotes(column.quote))
```

graciegoheen (approved):
Let's merge it ! Thanks for all the hard work here, I know folks are going to be stoked !

dave-connors-3 (approved):
neat

b-per (approved):

PR #395, opened 2023-11-20
add base models to accepted model types. 

### LINKED ISSUE #387: dbt_project_evaluator disagrees with dbt's own best practices guide for base models
I ran into a disagreement between `dbt_project_evaluator` and dbt's own [How we structure our dbt projects](https://docs.getdbt.com/guides/best-practices/how-we-structure/1-guide-overview) guide.

The guide includes two staging `base` models, inside the `models/staging/jaffle_shop/base` directory:

image

These models are named:
- `base_jaffle_shop__customers.sql`
- `base_jaffle_shop__deleted_customers.sql`

If you try this in your own project, `dbt_project_evaluator` will complain that these models should have a `stg_` prefix.

And similarly, `fct_staging_dependent_on_staging` would fail, claiming that `staging` models are depending on other `staging` models.

Given that both of these recommendations come from dbt... can we get some consistency?

dave-connors-3:
This is a:
- [ ] new functionality

## Link to Issue 
Closes #387 

## Description & motivation

Adds base models as an acceptable model type. 

Chose to exclude the new model i added to the integration tests so as not to have to edit all the test and documentation related seeds, but validated the output in `int_all_graph_resources`:

image

I also added an accepted values test to make sure the types match the variable -- not a perfect test, but at least protects against nulls and/or some faulty logic

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB
    - [ ] Trino/Starburst

### DIFF
--- a/integration_tests/dbt_project.yml
+++ b/integration_tests/dbt_project.yml
@@ -61,7 +61,7 @@ vars:
   # ensure integration tests run successfully when there are 0 of a given model type (extra)
   exclude_packages: ['exclude_package']
   exclude_paths_from_project: ["/to_exclude/","source_3.table_6"]
-  model_types: ['staging', 'intermediate', 'marts', 'other', 'extra', 'new_model_type']
+  model_types: ['base','staging', 'intermediate', 'marts', 'other', 'extra', 'new_model_type']
   # dummy variable used for testing fct_hard_coded_references
   my_table_reference: 'grace_table'
   new_model_type_folder_name: 'my_new_models'
--- a/integration_tests/models/dbt_project_evaluator_schema_tests/core.yml
+++ b/integration_tests/models/dbt_project_evaluator_schema_tests/core.yml
@@ -23,8 +23,10 @@ models:
           - is_not_empty_string
       - name: model_type
         tests: 
-          - accepted_values:
-              values: "{{ var('model_types') }}"
+          - dbt_utils.expression_is_true:
+              expression: "= 'base'"
+              config:
+                where: "resource_name = 'base_model_10'"
 
   - name: int_direct_relationships
     description: "This table shows one record for each direct parent/child pair in the graph."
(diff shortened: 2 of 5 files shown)


integration_tests/models/dbt_project_evaluator_schema_tests/core.yml
```diff
@@ -21,6 +21,10 @@ models:
         description: this contains the on_schema_change setting for incremental models. This column was sometimes an empty string, so should be tested to detect regressions
         tests:
           - is_not_empty_string
+      - name: model_type
+        tests: 
+          - accepted_values:
+              values: "{{ var('model_types') }}"
```
graciegoheen, line 27: Won't this also include 'extra' and 'new_model_type'? Is that desired?

integration_tests/dbt_project.yml
```diff
@@ -61,7 +61,7 @@ vars:
   # ensure integration tests run successfully when there are 0 of a given model type (extra)
   exclude_packages: ['exclude_package']
   exclude_paths_from_project: ["/to_exclude/","source_3.table_6"]
-  model_types: ['staging', 'intermediate', 'marts', 'other', 'extra', 'new_model_type']
+  model_types: ['base','staging', 'intermediate', 'marts', 'other', 'extra', 'new_model_type']
```
graciegoheen, line 64: This is just for the integration test right?

dave-connors-3: yep -- that's what happens when you commit from inside a subdirectory 🤦

graciegoheen:
> Chose to exclude the new model i added to the integration tests so as not to have to edit all the test and documentation related seeds, but validated the output in int_all_graph_resources

sounds like a good use case for unit testing.... hahaha

graciegoheen (approved):
LGTM

PR #416, opened 2024-01-24
Remove arg from `filter_exceptions`

b-per:
This is a:
- [x] new functionality

## Description & motivation

Half between bugfix and new functionality. This just simplifies how we filter the exceptions.

This is a breaking change for people who overwrite the definition of `filter_exceptions()` in their own project, so we should release it only in a new minor and not a patch.

It would be good to update the release notes and/or changelog mentioning the breaking change as well.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB
    - [ ] Trino/Starburst

### DIFF
--- a/macros/filter_exceptions.sql
+++ b/macros/filter_exceptions.sql
@@ -1,15 +1,15 @@
-{% macro filter_exceptions(model_name) -%}
-    {{ return(adapter.dispatch('filter_exceptions', 'dbt_project_evaluator')(model_name)) }}
+{% macro filter_exceptions() -%}
+    {{ return(adapter.dispatch('filter_exceptions', 'dbt_project_evaluator')()) }}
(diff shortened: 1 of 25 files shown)


graciegoheen:
What's the reason for this change? Just to simplify the macro? Changes look good!

b-per:
The main reason is that I only learned today that we could access `model` from a macro without having to pass it as an argument of the macro itself. And, yes, doing so makes it cleaner/easier to maintain. 

We actually had a bug previously with the macro not being defined properly in some new models.

graciegoheen (approved):

graciegoheen (approved):
alright

PR #419, opened 2024-02-01
[cleanup] replace wrap_string_with_quotes with dbt.string_literal

### LINKED ISSUE #418: replace wrap_string_with_quotes custom macro with dbt.string_literal()

graciegoheen:
Closes #418


b-per:
😬

graciegoheen:
@b-per i was really hoping this would "just work" :((((((

b-per:
We can't just replace our macro by `dbt.string_literal` because we handle nulls/empty strings differently.

I think that we should:
1. handle nulls differently still
2. but call `dbt.string_literal` when the value is not null

There is some benefit in knowing when a value is null or when it is the empty string (which is actually different). I will create a new PR for it.

Issue #421, opened 2024-02-04
Multiline column descriptions result in unclosed string literals


katieclaiborne:
### Describe the bug

In BigQuery, the new `base_<node/source>_columns` models fail with a syntax error when a multiline column description is present.

### Steps to reproduce

Take a single line description, and convert to a multiline description using block notation.
```yaml
description: >
  My column description
```
### Expected results

I expected dbt project evaluator to run without issue.

### Actual results

Instead, the new `base_<node/source>_columns` models failed with "Unclosed string literal" syntax errors.

### Screenshots and log output

Inspecting the failed BigQuery job shows that it is interpreting the multiline string as unclosed individual strings.

<img width="343" alt="image" src="https://github.com/dbt-labs/dbt-project-evaluator/assets/46941959/4c4ba882-de16-42f4-8d19-090bed5d037c">

### System information
**The contents of your `packages.yml` file:**
```yaml
packages:
  - package: dbt-labs/dbt_project_evaluator
    version: [">=0.9.0", "<0.10.0"]
```
**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [x] bigquery
- [ ] snowflake
- [ ] trino/starburst
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
I don't have ready access to the CLI at the moment, but these are the versions listed at the start of a Cloud IDE invocation.

dbt=1.6.9, dbt-bigquery=1.6.9

### Additional context

I think I've traced the issue to the [active line](https://github.com/dbt-labs/dbt-project-evaluator/blob/eb32a30d06fcf4011b043d1ef8aafccb60a2b5b6/macros/wrap_string_with_quotes.sql#L5) in the `wrap_string_with_quotes` macro.

It seems like you all are trying to move away from the macro in #419, but I don't know whether `dbt.string_literal` has a triple-quoting option to support [multiline BigQuery strings](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#string_and_bytes_literals).

### Are you interested in contributing the fix?

Yes! I've played around in #420, and have confirmed that it resolves the issue in our own project.

b-per:
Thanks for raising it!

We will fix it in a patch in the next day or so. 
Ideally, it also requires updates to dbt-bigquery which I did a PR for https://github.com/dbt-labs/dbt-bigquery/pull/1089 , but we can include the new `bigquery__string_literal` in this package for now.

[...]

PR #422, opened 2024-02-05
Use `dbt.string_literal` behind the scenes for quoting non-null values

b-per:
Alternative to #419 for wrapping strings with quotes

Add test for #421 which is failing for BQ without the fix


b-per:
I think that our test Redshift server is down and the latest databricks version is still not working with pandas >= 2.2

But the test passes on BigQuery! 😄

integration_tests/models/staging/source_1/schema.yml
```diff
@@ -5,7 +5,9 @@ models:
     description: test
     columns:
       - name: id 
-        description: hocus pocus
+        description: >
```
graciegoheen, line 8: lol

integration_tests/models/dbt_project_evaluator_schema_tests/core.yml
```diff
@@ -17,10 +17,6 @@ models:
         tests:
           - unique
           - not_null
-      - name: on_schema_change
```
graciegoheen, line 20: why remove this test?

macros/wrap_string_with_quotes.sql
```diff
@@ -1,7 +1,14 @@
...
+  {% if str is none %}
     {{ return('cast(NULL as ' ~ dbt.type_string() ~ ')') }}
   {% else %}
-    {{ return("'" ~ str ~ "'") }}
+    {{ dbt.string_literal(str) }}
   {% endif %}
 {% endmacro %}
+
+{#
+  To be removed when https://github.com/dbt-labs/dbt-bigquery/pull/1089 is merged
```
graciegoheen, line 10: Let's touch base with adapters team to see if this is something we can merge asap

models/marts/dag/fct_hard_coded_references.sql
```diff
@@ -11,7 +11,7 @@ final as (
         resource_name as model,
         hard_coded_references
     from models
-    where hard_coded_references is not null
+    where hard_coded_references != ''
```
graciegoheen, line 14: why the change from null to empty string?

b-per: Because I changed `wrap_string_with_quotes` to handle `none` and `''` separately

integration_tests/models/dbt_project_evaluator_schema_tests/core.yml
b-per: Because it failed 🤣 

If I recall, this is because we now handle empty strings or None differently in `wrap_string_with_quotes(str)`

b-per:
🤔 Why is this failing on `fct_test_directories` now...

b-per:
OK, got it. 

A side effect of null vs empty string handling.

dave-connors-3 (approved):

PR #443, opened 2024-04-04
Avoid SQL line comments when searching hard coded tables

b-per:
This is a:
- [ ] new functionality
- [x] minor improvement to the current behavior

## Link to Issue 

No issue, but came from [this internal Slack thread](https://dbt-labs.slack.com/archives/C02TRN1757B/p1712242266111429)

## Description & motivation

This enhances our approach to search for hard coded tables, removing the inline comments (by default `--` but with a new variable to update it)

Being regex, this is not perfect. For example, now we won't find the hard coded references if someone has the following code

```
select '--whatever' from myschema.mytable
```

(e.g. when `--` is not part of a comment)

But I think that this happens much less than people commenting their SQL with `--`

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [x] DuckDB
    - [ ] Trino/Starburst

### DIFF
--- a/integration_tests/models/marts/fct_model_6.sql
+++ b/integration_tests/models/marts/fct_model_6.sql
@@ -5,86 +5,89 @@
 }}
 
 select 1 as id 
--- from {{ ref('stg_model_3') }}
 
--- union all 
--- select
---     3 as id
--- from my_db.my_schema.my_table
--- union all 
--- select
---     3 as id
--- from 'my_db'.'my_schema'.'my_table'
--- union all 
--- select
---     3 as id
--- from "my_db"."my_schema"."my_table"
--- union all 
--- select
---     3 as id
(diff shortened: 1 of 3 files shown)


integration_tests/models/marts/fct_model_6.sql
```diff
@@ -5,86 +5,89 @@
...
+  select
+      4 as id
+  from {{ var("my_table_reference") }}
+  union all
+  select
+      4 as id
+  from {{ var('my_table_reference') }}   
+  union all
+  
+  -- the following is SQL commented so it should be found as a hard coded reference
```
dave-connors-3, line 79: so it should _not_ be found, right?

dave-connors-3 (approved):
one comment on your comment that i don't think is blocking!

PR #452, opened 2024-05-10
Incorporate not null constraint into primary key rule

### LINKED ISSUE #373: Support not-null constraints as part of primary key definition
### Describe the feature
I'd like for a model with a unique test and a not-null constraint applied to the same column to satisfy the [primary key test](https://dbt-labs.github.io/dbt-project-evaluator/0.7/rules/testing/#missing-primary-key-tests) rule.

In our project, we've considered overriding the `primary_key_test_macros` variables to define a uniqueness test alone as valid. This would allow developers to replace not-null tests with constraints, but it wouldn't validate that they have.

```
    primary_key_test_macros: [
      ["data_platform.test_unique"],
      ["data_platform.test_unique_combination_of_columns"]
    ]
```

### Additional context
We use BigQuery, so the not-null constraint is the only one available to us.

### Who will this benefit?
Teams looking to leverage constraints without duplicating existing tests or sacrificing test coverage.

I'd love to contribute, but would need a hand getting started.

I've poked around a bit to understand how the `primary_key_test_macros` variable is currently used, and I've inspected how constraints appear in the project manifest. The tricky bit seems to be that column-level constraints appear as part of the relevant model node, while tests appear as their own nodes, with dependencies on (test) macro and model nodes.

katieclaiborne7:
This is a:
- [x] new functionality

## Link to Issue 
Closes #373

## Description & motivation

These changes add a third default method of defining a model's primary key: a not null constraint and unique test applied to the same column.

Expanding the default primary key definitions will enable teams to leverage constraints without duplicating existing tests or sacrificing test coverage.

image

## Checklist
    - [x] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB
    - [ ] Trino/Starburst


b-per:
Thanks @katieclaiborne7 !
Is it still Draft/WIP, or is it ready to be reviewed/merged?

katieclaiborne7:
Ready!

models/marts/tests/intermediate/int_model_test_summary.sql
```diff
@@ -32,6 +38,34 @@ count_column_tests as (
     group by 1,2
 ),
 
+count_column_constraints as (
+
+    select
+        node_unique_id as direct_parent_id,
+        name as column_name,
+        case
+            when replace(constraints, 'not_null', '') != constraints
```
dave-connors-3, line 47: not a huge deal, but worth calling out that this approach means we'd count anything that matches this string like `not_not_null` or `not_nulls` as a not_null constraint -- probably pretty unlikely given what kind of constraints exist out there in the wild, but worth calling out if someone uses the [expression option](https://docs.getdbt.com/reference/resource-properties/constraints) and does something related to not null we may get false positives.

Maybe we could check for `"type": "not_null"` or simply `"not_null"`?

models/marts/tests/intermediate/int_model_test_summary.sql
```diff
@@ -59,7 +95,8 @@ final as (
         all_graph_resources.resource_type,
         all_graph_resources.model_type,
         coalesce(agg_test_relationships.is_primary_key_tested, FALSE) as is_primary_key_tested,
-        coalesce(agg_test_relationships.number_of_tests_on_model, 0) as number_of_tests_on_model
+        coalesce(agg_test_relationships.number_of_tests_on_model, 0) as number_of_tests_on_model,
+        coalesce(agg_test_relationships.number_of_constraints_on_model, 0) as number_of_constraints_on_model
```
dave-connors-3, line 99: for consistency's sake, we may want to full this through to the `fct_` model as well! not blocking

dave-connors-3 (approved):
some small, non-blocking feedback!

models/marts/tests/intermediate/int_model_test_summary.sql
katieclaiborne7: Good call! I've updated the line to check for `"not_null"`.

models/marts/tests/intermediate/int_model_test_summary.sql
katieclaiborne7: Hmm, which `fct_` model are you thinking of? I believe the column already makes its way into `fct_missing_primary_key_tests`, courtesy of a few `select *` statements.

katieclaiborne7 (commented):
Thanks for the feedback, Dave!

models/marts/tests/intermediate/int_model_test_summary.sql
```diff
@@ -32,6 +38,34 @@ count_column_tests as (
...
+
+    select
+        node_unique_id as direct_parent_id,
+        name as column_name,
+        case
+            when replace(constraints, '"not_null"', '') != constraints
+            then 1
+            else 0
+        end as constraint_not_null_count,
+        cast((length(constraints) - length(replace(constraints, 'type', ''))) / 4 as int) as constraints_count
```
b-per, line 51: Could we move this logic to Jinja in `base_node_columns.sql`/`get_column_values.sql`?

We can keep `constraints` as a column if people want to drill into the constraints but extracting the ones we are interested in seems better in Jinja rather than doing some clever string operations.

We could give them the names `has_not_null_constraint` and `constraints_count` for example.

katieclaiborne7: I've given this a try! Let me know what you think.

b-per (approved):
Thanks @katieclaiborne7

PR #453, opened 2024-05-10
Add config for materialization

jschintz-nytimes:
This is a:
- [X ] new functionality

## Description & motivation
Add aditional fields, specific to NYT deployment, describing the dbt graph nodes.

Rushed - its friday after 5


dave-connors-3:
hey @jschintz-nytimes hope you had a nice weekend :) 

do you want to open an issue here? happy to chat about adding these fields!

jschintz-nytimes:
Hiya @dave-connors-3.  
No need - I just wasn't paying attention when I opened a PR for my fork.

But, the use case if think its useful.
dbt-bigquery has some pretty useful config options that can decrease $ spend & governance in general.  Its hard to get all engineers familiar with the adapter configs.

Our deployment is considering applying some opinionated rulesets that:
- copy_partitions should be true for any incremental model where the field type is a date and is >1MM rows.  
- incremental_strategy should not be 'merge' for any table over 1MM rows.  Too expensive.
- fullrefresh should not be true for any table over 100MM rows.
- persist_docs should be true on all models.  Documentation should live everywhere the data is.

These are adapter specific - so I'd get that this might not be for something general like project evaluator.  

I saw the product demo yesterday for the new UX around CI, data changes & recommendations. Really looking forward to that!

b-per:
@dave-connors-3 , we could maybe add `config` as a raw JSON object and let people do whatever they want with it.

That way it would work for all adapters and would still allow people to unnest the JSON objects.

PR #475, opened 2024-07-03
Type string macro handling all cases

### LINKED ISSUE #469: redshift__type_string overwrite is causing all existing models to be marked as modified
### Describe the bug
I configured dbt_project_evaluator with dispatch required for redshift. My intention is to use it in a slim CI. 
I have a job in DBT Cloud which defers to the production environment. After configuring dispatch all models are recognised as modified. 

### Steps to reproduce
1. Configure dbt_project_evaluator package with dispatch in existing redshift project
2. Run 
```
dbt ls --select state:modified+ --exclude package:dbt_project_evaluator
``` 
4. See that models changed

### Expected results
It should be implemented in a way that does not overwrite core macros for all existing models. 

### Actual results
After package configuration I expect no models to be marked as changed. 

### System information
```

dispatch:
  - macro_namespace: dbt
    search_order: ['dbt_project_evaluator','dbt']
```

- [ ] postgres
- [x] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] trino/starburst

```
dbt Cloud CLI - 0.38.4 (0e2097036f895b2f615c53e144e625f88c8f9012 2024-06-21T16:58:57Z)
```
DBT version 
12:26:46  Running with dbt=1.6.16
12:26:47  Registered adapter: redshift=1.6.7

### Additional context
https://github.com/dbt-labs/dbt-project-evaluator/blob/main/macros/cross_db_shim/redshift_shims.sql

### Are you interested in contributing the fix?
I can contribute, however I'm not sure how to fix it at this point.

mweso-softserve:
Creates a macro that redefine type_string for models in dbt_project_evaluator package. No need to override it for all models in a project anymore when using package.
This attempt uses api.Column.string_type(600) for all databases except bigquery. 

This is a:
- [ ] new functionality

## Link to Issue 
 
Closes #469 

## Description & motivation
This implementation replaces `type_string()` macro which does't work for redshift for tables defined in the way it was implemented in dbt-project-evaluastor with `api.Column.string_type(600)`

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [x] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB
    - [ ] Trino/Starburst

I tested that on a local version of this package without dispatcher configuration.


b-per:
I had a chat with the team and we don't really want to introduce a `dbt_project_evaluator` version of `type_string()`.

We understand that with https://github.com/dbt-labs/dbt-project-evaluator/issues/469, the first time CI when the package is added it might pick up more models than needed, but this should happen only once and could be handled manually.

I am going on leave for a couple of weeks and will want to revisit the few different issues we have around strings with Redshift but this particular PR is likely not one we would want to merge to this repo.

mweso-softserve:
@b-per 

> I had a chat with the team and we don't really want to introduce a `dbt_project_evaluator` version of `type_string()`.
> 
> We understand that with #469, the first time CI when the package is added it might pick up more models than needed, but this should happen only once and could be handled manually.
> 
> I am going on leave for a couple of weeks and will want to revisit the few different issues we have around strings with Redshift but this particular PR is likely not one we would want to merge to this repo.

`dbt_project_evaluator` [already introduced](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/macros/cross_db_shim/redshift_shims.sql) its version of `type_string()` however the way it did changes the type definition for all redshift models. I'm not saying this PR is definitely the way to go, but the benefit is it redefines `type_string()` for `dbt_project_evaluator's` models only not the entire project. Let's imagine I want to use two different dbt packages that handle this kind of problem in two different ways. Both would try to redefine `type_string()` for the entire project and depending on the dispatch config one of them would always win potentially braking functionality of the other.  

It's not only the issue of dealing with models being marked as changed the first time it runs, but redefining ultimately all existing models whenever the string type was used. 
I don't agree that any package should modify definitions of types in exiting models simply by adding its configuration, unless that's the sole purpose of the package. 
Please revisit the approach.

glsdown:
I have to agree with @mweso-softserve here. In your current approach you have overwritten a core macro `type_string` used across every model in every Redshift dbt project. This instead will namespace that usage so it only becomes applicable to project evaluator models.

b-per:
Thanks for the feedback. I was out of office for a bit but I will try to talk again with the team by next week.

b-per:
We had a chat internally today.

As both of you raise the same point, we are going forward with this approach, creating a `dbt_project_evaluator.type_string()` macro. 

Hopefully it should help people loading strings in Redshift which we a few issues were raised about.

dave-connors-3 (approved):

Issue #496, opened 2024-09-18
Duplicate tests found when overriding default dbt.test_unique


PatFitzner:
### Describe the bug

If the `unique` test is overriden within a project, it follows that the variable `primary_key_test_macros` should be explicitly declared in the project.yml, specifying the project's own test version instead of the builtin (because our test will otherwise not be considered to cover unique key testing).

Since `int_all_graph_resources` is composing a list of tests from the `primary_key_test_macros` variable, our version of `unique` will be included.

However, right after the test list is composed, the line 
``` yaml
{%- do test_macro_list.append("dbt.test_unique") -%}
```
adds the builtin version of `unique` to the list, which leads the block 
``` yaml
{% for test in test_macro_set %}
        unioned_with_calc.macro_dependencies like '%macro.{{ test }}%' and unioned_with_calc.resource_type = 'test' as is_{{ test.split('.')[1] }}
```
to cause a name collision, since both `our_project.test_unique` and `dbt.test_unique` return the same value after `test.split('.')[1]`.
 

### Steps to reproduce
In a dbt project (let's use [jaffle-shop-classic](https://github.com/dbt-labs/jaffle-shop-classic) as an example), install the **dbt_project_evaluator** package, and run it 

``` bash
dbt run -s package:dbt_project_evaluator 
```
to ensure that it is working properly.

Create a custom version of the `unique` test as `jaffle-shop-classic/tests/test_unique.sql`, such as:
``` yaml
{% test unique(model, column_name) %}

{{ config(severity = 'warn') }}

select
    {{ column_name }} as unique_field,
    count(*) as n_records
from 
    {{ model }}
where 
    `{{ column_name }}` is not null
group by 
    `{{ column_name }}`
having 
    count(*) > 1

{% endtest %}
```
and add the variable 
``` yaml
primary_key_test_macros: [["jaffle_shop.test_unique", "dbt.test_not_null"],["dbt_utils.test_unique_combination_of_columns"]]
```
to your `dbt_project.yml` .

Run
``` bash
dbt run -s package:dbt_project_evaluator 
```

### Expected results
The package builds its models and the results are leveraging our custom test instead of the built-in one.

### Actual results
We receive a 
``` bash
Database Error in model int_all_graph_resources (models/marts/core/int_all_graph_resources.sql)
  column "is_test_unique" specified more than once
```
because `int_all_graph_resources` is appending the builtin `unique` test to the `test_macro_list`, causing a name collision with our test.

[...]

PR #497, opened 2024-09-19
Built-in unique test to not be explicitly added to "primary_key_test_macros"

PatFitzner:
In order to avoid name collisions on unique test definition overrides, the line that explicitly adds the builtin `dbt.test_unique` test to the "primary_key_test_macros" list has been replaced with a conditional block that will only add it if none of the tests already present in the "primary_key_test_macros" list would produce a `is_test_unique` column in the `int_all_graph_resources` model, which is required downstream by `int_model_test_summary`.

This is a:

## Link to Issue 

Closes [496](https://github.com/dbt-labs/dbt-project-evaluator/issues/496)

## Description & motivation

Line 10 of `models/marts/core/int_all_graph_resources.sql` 
```yaml
{%- do test_macro_list.append("dbt.test_unique") -%}
```
Is supposed to ensure that the column `is_test_unique` is generated, but breaks the execution of the package when the default test `dbt.test_unique` has been overridden in a project.

## Checklist
    - [x] BigQuery
    - [x] Postgres


PatFitzner:
I feel like CircleCI is out to get me. I've pushed the same commit 3 or 4 times with amended comments to trigger a new CI execution, and the failures are not always the same, as well as unrelated to the changes of this PR (non-exsiting tables, failed freshness / non-documented model tests ...)
Could you please let me know what I may be missing here @graciegoheen @dave-connors-3 @b-per ?

Thanks <3

b-per:
Thanks @PatFitzner 
Yes, CI is a bit flaky with some adapters unfortunately but we will have a look.

I think that the original piece of code on `test_unique` was added because we allow people to say that a model is PK-tested if it has a unique test and has a not null constraint on it.

I think that your approach is correct but I will need to do a few more checks.

b-per (approved):

PR #528, opened 2025-01-28
update gitignore to not include local duckdb instances

dave-connors-3:
plz


graciegoheen:
we mergin' this in or what?

PR #537, opened 2025-05-27
Move source freshness under config

graciegoheen:
Starting in 1.10, dbt Core is issuing deprecating warnings for moving `freshness` under `config` ([issue](https://github.com/dbt-labs/dbt-core/issues/11506)).

We are backporting the ability to set `freshness` under `config` to 1.9 ([issue](https://github.com/dbt-labs/dbt-core/issues/11681)).

This PR:
- moves source `freshness` under config
- updates required dbt version to 1.9

### DIFF
--- a/macros/unpack/get_source_values.sql
+++ b/macros/unpack/get_source_values.sql
@@ -24,8 +24,8 @@
               "cast(" ~ dbt_project_evaluator.is_not_empty_string(node.description) | trim ~ " as " ~ dbt.type_boolean() ~ ")",
               "cast(" ~ node.config.enabled ~ " as " ~ dbt.type_boolean() ~ ")",
               wrap_string_with_quotes(node.loaded_at_field | replace("'", "_")),
-              "cast(" ~ ((node.freshness != None) and (dbt_project_evaluator.is_not_empty_string(node.freshness.warn_after.count) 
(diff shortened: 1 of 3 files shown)


macros/unpack/get_source_values.sql
```diff
@@ -24,8 +24,8 @@
               "cast(" ~ dbt_project_evaluator.is_not_empty_string(node.description) | trim ~ " as " ~ dbt.type_boolean() ~ ")",
               "cast(" ~ node.config.enabled ~ " as " ~ dbt.type_boolean() ~ ")",
               wrap_string_with_quotes(node.loaded_at_field | replace("'", "_")),
-              "cast(" ~ ((node.freshness != None) and (dbt_project_evaluator.is_not_empty_string(node.freshness.warn_after.count) 
-                or dbt_project_evaluator.is_not_empty_string(node.freshness.error_after.count))) | trim ~ " as boolean)",
+              "cast(" ~ ((node.config.freshness != None) and (dbt_project_evaluator.is_not_empty_string(node.config.freshness.warn_after.count)
```
graciegoheen, line 27: actually, i'm realizing this logic should probably be:
```
node.config.freshness

OR 

node.freshness
```

before `node.freshness` is actually fully deprecated 

and same for below

b-per (approved):

PR #580, opened 2026-06-02
fix: add --no-manage-state to Fusion test runs

b-per:
## Summary

- Adds `--no-manage-state` flag to all `dbt build`/`dbt run` commands in `run_fusion_tests.sh`
- This flag skips source freshness state management, which is not yet implemented in the DuckDB Fusion adapter (causes a panic: `not yet implemented: DuckDBAdapter::freshness`)
- Verified locally: all three integration test projects (integration_tests, integration_tests_2) pass cleanly with this flag on dbt Fusion + DuckDB

## Test plan

- [ ] `run-fusion-duckdb-tests` CI job passes

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -85,7 +85,7 @@ vars:
   other_prefixes: ['rpt_']
 
   # -- Performance variables --
-  chained_views_threshold: "{{ 5 if target.type not in ['athena', 'trino'] else 4 }}"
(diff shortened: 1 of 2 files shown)


joellabes:
This is OK for the sake of unblocking yourself, but it shouldn't be necessary as a long term solution ([Slack thread](https://dbt-labs.slack.com/archives/C08AA468V54/p1780440528326739?thread_ts=1780410676.759209&cid=C08AA468V54))

b-per:
OK. I might not merge it then.
Anyway, project-evaluator CI is broken in Fusion at the moment 🤣

PR #583, opened 2026-06-12
fix: use valid macro argument type for loop_vars vars

ota2000:
This is a:
- [x] bug fix PR with no breaking changes
- [ ] new functionality

## Link to Issue

Closes #582

## Description & motivation

dbt-core 1.11 validates macro argument `type` annotations against a fixed type grammar (`macro_types` in `dbt/parser/schemas.py`). Union syntax such as `list|string` is not part of that grammar, so every `dbt parse` / `compile` / `run` against a project installing this package emits:

```
Argument vars in the yaml for macro loop_vars has an invalid type.
```

This PR changes the annotation to `type: list[string]`, which reflects what the argument actually is: callers (`stg_naming_convention_folders.sql` / `stg_naming_convention_prefixes.sql`) build a list of variable *names* and the macro iterates it with `for var_name in vars`. The list-or-string union in the old annotation described the dbt var *values* looked up inside the macro (`var(var_name)` can be a string or a list), not the argument itself — so no expressiveness is lost. The description is updated accordingly.

I scanned all macro yml files in the package with dbt-core's `is_valid_type` and this was the only invalid annotation.

Verified locally with dbt-core 1.11.11: the warning disappears after this change and `dbt parse` succeeds.

## Integration Test Screenshot

Not applicable — this is a yml metadata change (macro argument annotation only); no model or macro logic is touched, so integration test behavior is unchanged.

## Checklist
- [x] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [x] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] Databricks
    - [ ] DuckDB
    - [ ] Trino/Starburst
- [ ] I have updated the README.md (if applicable) — not applicable
- [ ] I have added tests & descriptions to my models (and macros if applicable) — not applicable (annotation-only change; argument description updated)

PR #584, opened 2026-06-16
fix: use valid type annotation for loop_vars macro argument

### LINKED ISSUE #582: dbt-core 1.11 warns: "Argument vars in the yaml for macro loop_vars has an invalid type"
### Describe the bug

Starting with dbt-core 1.11, every `dbt parse` / `dbt compile` / `dbt run` against a project that installs this package emits the following warning:

```
Argument vars in the yaml for macro loop_vars has an invalid type.
```

dbt-core 1.11 validates macro argument `type` annotations against a fixed type grammar ([`macro_types` in `dbt/parser/schemas.py`](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/parser/schemas.py)): valid names are `str`, `string`, `bool`, `int`, `integer`, `float`, `any`, `list[...]`, `dict[...,...]`, `optional[...]`, `relation`, `column`. Union syntax such as `list|string` is not part of the grammar, so the annotation in [`models/staging/variables/variables.yml`](https://github.com/dbt-labs/dbt-project-evaluator/blob/main/models/staging/variables/variables.yml) fails validation:

```yaml
macros:
  - name: loop_vars
    arguments:
      - name: vars
        type: list|string   # <- not a valid type expression for dbt-core >= 1.11
```

This is the only invalid type annotation in the package (I scanned all macro yml files with dbt-core's `is_valid_type`).

### Steps to reproduce

1. Install `dbt-labs/dbt_project_evaluator` (reproduced on 1.2.4; the annotation is unchanged on v1.3.0 and `main`)
2. Run `dbt parse` with dbt-core 1.11.x

### Expected results

No warning.

### Actual results

```
02:34:08  Argument vars in the yaml for macro loop_vars has an invalid type.
```

emitted on every invocation (it also shows up in CI logs for all downstream projects).

### Proposed fix

Change the annotation to `type: any` — the macro accepts either a list or a string, and `any` is the only way to express that union in the dbt-core 1.11 type grammar. Verified locally that this makes the warning disappear. I can open a PR.

### System information

```yaml
packages:
  - package: dbt-labs/dbt_project_evaluator
    version: 1.2.4
```

- [x] bigquery (the warning is warehouse-agnostic; it fires at parse time)

```
Core:
  - installed: 1.11.11
```

b-per:
## Summary

- Changes `type: list|string` to `type: any` in `models/staging/variables/variables.yml` for the `loop_vars` macro's `vars` argument
- dbt-core 1.11 validates macro argument type annotations against a fixed grammar (`str`, `string`, `bool`, `int`, `float`, `any`, `list[...]`, `dict[...]`, `optional[...]`, `relation`, `column`) — union syntax like `list|string` is not valid and triggers a warning on every invocation
- `any` is the correct annotation for an argument that accepts either a list or a string

Fixes #582

### DIFF
--- a/models/staging/variables/variables.yml
+++ b/models/staging/variables/variables.yml
@@ -5,5 +5,5 @@ macros:
     description: A macro that loops through variables and returns them as a SQL query to be used in a model
     arguments:
       - name: vars
-        type: list|string
+        type: any
         description: A list of variables from dbt_project.yml


graciegoheen:
@b-per do we want to set `validate_macro_args: true` in dbt_project.yml too?

dave-connors-3:
hey folks

b-per:
Closing in favour of #583 (community-authored, rebased on main) which uses the more precise `list[string]` type. The `validate_macro_args: true` flag for integration tests is being added on top of that work.

b-per:
> hey folks

👋

PR #588, opened 2026-06-24
fix: escape single quotes in metric type_params.expr before INSERT

b-per:
## Summary

- `get_metric_values()` passed `node.type_params.expr` directly to `wrap_string_with_quotes()`, which calls `dbt.string_literal()` and wraps the value in single quotes
- When `type_params.expr` contains a single quote (e.g. a CASE expression like `case when id = 'active' then 1 else 0 end` from a measure converted by dbt Wizard), the generated INSERT SQL is syntactically invalid
- Fix: wrap with `dbt.escape_single_quotes()` before passing to `wrap_string_with_quotes()`, matching the treatment already applied to `label` and `filter`

## Regression test

Added `metric_with_expr_single_quote` to `integration_tests_sl` with `expr: "case when id = 'active' then 1 else 0 end"`. Without the fix this produces:

```
Parser Error: syntax error at or near "active"
LINE 58: 'case when id = 'active' then 1 else 0 end'
```

## Related

Reported via ZD-125778 (Frog Street)

### DIFF
--- a/integration_tests_sl/models/marts/intermediate/_dim_model_7.yml
+++ b/integration_tests_sl/models/marts/intermediate/_dim_model_7.yml
@@ -47,3 +47,9 @@ models:
           {{ Dimension('dim__id') }} > 0
         window: 30 days
         input_metric: total_count
+      - name: metric_with_expr_single_quote
+        label: "Metric with single quote in expr"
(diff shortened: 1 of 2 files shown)


b-per:
CI passing

PR #594, opened 2026-08-25
feat: publish v2 native project checks

dataders:
> [!NOTE]
> This is work-in-progress spike work to identify what is needed to make dbt-project-evaluator compatible with the dbt information schema and dbt checks. It is not yet a release-ready implementation.

## Summary

- explore publishing the native-check implementation as dbt-project-evaluator `2.0.0`
- keep the package `dbt_project.yml` valid in dbt Core by moving `info_schema` and check defaults to `checks/_checks.yml`
- leave installed native checks disabled until the consuming project opts in
- document package-wide and per-check configuration, including severity overrides
- retain the legacy warehouse models for dbt Core users
- add a Fusion-only consumer fixture that installs the package and runs all 28 supported native checks

## Consumer opt-in

```yaml
models:
  dbt_project_evaluator:
    +enabled: false

checks:
  dbt_project_evaluator:
    +enabled: true
```

Selective opt-in is also supported by setting the package to `+enabled: false` and enabling named checks beneath it.

## Coverage

This spike implements 28 of the 29 evaluator rules. `fct_hard_coded_references` remains deferred because it requires parsed SQL or lint/static-analysis findings rather than the parse-time information schema.

## Native PR stack

1. #593
2. #594 (this PR)

The stack also depends on dbt-labs/fs#13654, which is stacked on dbt-labs/fs#13598.

## Verification

- freshly rebuilt debug Fusion binary on the framework branch
- installed-package DuckDB smoke test: all 28 checks executed; 25 passed and 3 produced advisory warnings; command succeeded
- dbt Core 1.12.3 dependency load and parse succeeded with evaluator v2 installed
- `bash -n run_fusion_tests.sh`
- one metadata entry exists for each of the 28 check SQL files

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,28 @@ Currently, the following adapters are supported:
 
 ## Using This Package
 
+### Fusion native checks (v2)
+
+Version 2 includes native, parse-time project checks for 28 of the package's 29
+rules. Installing the package does not enable them automatically. Fusion users
+can disable the legacy warehouse models and opt into all native checks from the
+root project's `dbt_project.yml`:
+
+```yml
+models:
+  dbt_project_evaluator:
+    +enabled: false
+
+checks:
+  dbt_project_evaluator:
+    +enabled: true
+```
+
+The checks are advisory by default. Root-project configuration can enable one
+rule at a time or override `severity`; see [the native-check guide](checks/README.md)
+for package-wide and selective examples. The existing model-based implementation
+remains available for dbt Core projects.
+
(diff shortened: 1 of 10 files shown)


b-per:
Can we use native GH PR stacks for this?

dataders:
Done — #594 now uses #593’s head branch as its base, so GitHub tracks the two PRs as a native stack. I also updated the PR body with stack navigation and the WIP spike framing.

